PR #3, opened 2017-02-09
Splitted the timed_api app into three apps.

anehx:


sliverc:
Nice and clean structure... let's merge it.

PR #5, opened 2017-02-24
Employments

anehx:
* Added locations
* Added public holidays
* Added employments
* Simplified API
* Changed to readonly endpoints for most of the objects
* Added some admin views


timed/settings.py
```diff
@@ -158,12 +143,12 @@ def trueish(value):
     ),
     'DEFAULT_PARSER_CLASSES': (
         'rest_framework_json_api.parsers.JSONParser',
+        'rest_framework.parsers.JSONParser',
         'rest_framework.parsers.FormParser',
         'rest_framework.parsers.MultiPartParser'
     ),
     'DEFAULT_PERMISSION_CLASSES': (
         'rest_framework.permissions.IsAuthenticated',
-        'rest_framework.permissions.DjangoModelPermissions',
```
sliverc: How are you going to handle permissions on different models after you have removed this?
All authenticated user can access all end points?

anehx: At the moment we don't need any complex permission, since we only have readonly models or models which can only be edited / deleted / created by their owner. It is not in sight that we need more complex permission but if we do, we can still add it back.

The removal of this permission class make testing much easier, since we don't have to add all the permissions to the users.

sliverc: Make sense...

sliverc:
Looks good... merging this.

Another thing I discovered while checking for the permissions. Doesn't have directly to do with this PR.

The current user is set in the view overwriting perform_create. Using CurrentUserDefault would be the cleaner approach to this.
See http://www.django-rest-framework.org/api-guide/validators/#currentuserdefault

PR #6, opened 2017-02-27
Better validations

anehx:
* Validate employments on model (only used in admin)
* Validate activity blocks on serializer

### DIFF
--- a/timed/employment/admin.py
+++ b/timed/employment/admin.py
@@ -5,6 +5,7 @@ import datetime
 from django import forms
 from django.contrib import admin
 from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
 
 from timed.employment import models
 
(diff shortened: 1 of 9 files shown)


timed/employment/models.py
```diff
@@ -39,6 +42,31 @@ class Employment(models.Model):
     start_date       = models.DateField()
     end_date         = models.DateField(blank=True, null=True)
 
+    def clean(self):
+        """Validate the employment as a whole.
```
sliverc: As http://www.django-rest-framework.org/topics/3.0-announcement/#differences-between-modelserializer-validation-and-modelform describes rest framework doesn't call clean and expects validation to happen on the Serializer. Only Forms call clean.

As this code base exposes a rest api but on the other hand also does server side rendering (django admin) it can get fairly confusing to do the validation on the model level but the api actually doesn't use this validation.

As a cleaner approach I therefore recommend to move the django admin validation to the Form.clean method of the django admin which needs it.
See here how to overwrite form: https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.ModelAdmin.form

timed/employment/admin.py
```diff
@@ -1,10 +1,66 @@
...
+        """
+        data = super().clean()
+
+        employments = models.Employment.objects.filter(user=data.get('user'))
+
+        if (
+            data.get('end_date') and
+            data.get('start_date') >= data.get('end_date')
+        ):
+            raise ValidationError('The end date must be after the start date')
```
sliverc: This is the first string introduced in the timed backend... even if timed is not internationalized at the time wouldn't it make sense to use the translate module here?

timed/employment/admin.py
```diff
@@ -1,10 +1,66 @@
...
+            employments.filter(end_date__isnull=True) and
+            data.get('end_date') is None
+        ):
+            raise ValidationError('A user can only have one active employment')
+
+        if any([
+            e.start_date <= (
+                data.get('end_date') if
+                data.get('end_date') else
+                datetime.date.today()
```
sliverc: Much easier like this:
`date.get('end_data') or datetime.date.today()`

timed/employment/admin.py
```diff
@@ -1,10 +1,66 @@
...
+            ) and data.get('start_date') <= (
+                e.end_date if e.end_date else datetime.date.today()
```
sliverc: easier way:
`e.end_date or datetime.date.today()`

PR #7, opened 2017-03-24
IPA

anehx:


timed/employment/serializers.py
```diff
@@ -82,3 +178,120 @@ class Meta:
...
+        fields = ['name']
+
+
+class AbsenceCreditSerializer(ModelSerializer):
+    """Absence credit serializer."""
+
+    absence_type = ResourceRelatedField(read_only=True)
+    user         = ResourceRelatedField(read_only=True)
+    used         = SerializerMethodField()
+    balance        = SerializerMethodField()
```
sliverc: if aligning is used this should be properly aligned as well I guess... :)

PR #9, opened 2017-05-01
Fixed issue where the generated report got deleted when the activity was deleted

anehx:


sliverc:
Looks good... as you have fixed quite a few doc strings here could you also fix the raised points in https://github.com/adfinis-sygroup/timed-backend/pull/7? Best would actually to separate those changes into 2 PRs but I guess not worth the effort for this.

anehx:
@sliverc done!

sliverc:
LGT.

PR #11, opened 2017-05-05
Added some filters for the weekly overview

anehx:


timed/tracking/filters.py
```diff
@@ -109,6 +109,9 @@ class Meta:
 class ReportFilterSet(FilterSet):
     """Filter set for the reports endpoint."""
 
+    from_date = DateFilter(name='date', lookup_expr='gte')
+    to_date   = DateFilter(name='date', lookup_expr='lte')
```
sliverc: I think it is good practice to add new filter to Meta fields as well.

PR #13, opened 2017-05-08
Added own model for absences

anehx:


timed/employment/serializers.py
```diff
@@ -91,7 +91,21 @@ def get_worktime_balance_raw(self, instance):
...
+                user=instance,
+                date__gte=start_date,
+                date__lte=end_date
+            ).values_list('duration', flat=True),
+            timedelta()
+        )
+
+        return (
+            reported_worktime +
+            absences +
```
sliverc: I think the absences should be documented in the docstring of this method as well.

timed/employment/models.py
```diff
@@ -39,6 +39,28 @@ class Employment(models.Model):
     start_date       = models.DateField()
     end_date         = models.DateField(blank=True, null=True)
 
+    @classmethod
+    def employment_at(cls, user, date):
```
sliverc: the preferred way to add table-level functionality is model Managers - see https://docs.djangoproject.com/en/1.11/topics/db/managers/#adding-extra-manager-methods

timed/employment/models.py
```diff
@@ -39,6 +39,28 @@ class Employment(models.Model):
...
+        try:
+            return cls.objects.get(
+                (
+                    models.Q(end_date__gte=date) |
+                    models.Q(end_date__isnull=True)
+                ),
+                start_date__lte=date,
+                user=user
+            )
+        except Exception:
```
sliverc: In general generic exceptions should not be caught or thrown...
better I guess here would be to catch Employment.DoesNotExist and raise ParseError

sliverc:
Nice... Merging.

PR #17, opened 2017-05-15
Added option to configure working days per location

sliverc:
* Per default Monday to Friday is selected.
* Uses workday information to calculate PublicHolidays
* This only works once PR #15 is merged


sliverc:
@anehx Added bugfix for PublicHolidays worktime balance issue as well

PR #22, opened 2017-06-29
Feature recent filters

kaldras:
Add "recent" filter for activities and reports. This allows to get the 10 most recent entries of the respective object.

### DIFF
--- a/timed/tracking/filters.py
+++ b/timed/tracking/filters.py
@@ -72,7 +72,7 @@ class ActivityActiveFilter(Filter):
 
 
 class RecentActivityFilter(Filter):
-    """Filter to get recent activities"""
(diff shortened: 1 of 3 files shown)


timed/tracking/filters.py
```diff
@@ -71,17 +71,33 @@ def filter(self, qs, value):
         ).distinct()
 
 
+class RecentActivityFilter(Filter):
+    """Filter to get recent activities."""
+
+    @boolean_filter
```
sliverc: would it be better to use an integer filter so the amount of recent items can actually be given?

Would be more flexible also for features where a recent item settings can be set by user.

kaldras:
This was redeemed by #24 - closing

PR #24, opened 2017-07-06
Added "my_most_frequent" filter to task end point

sliverc:
It only counts most frequent tasks of user which have been reported to in the last few months as only those are relevant for usability. Filter expects an integer of how may tasks need to be given.

This supersedes #22.

### DIFF
--- a/timed/projects/filters.py
+++ b/timed/projects/filters.py
@@ -44,7 +44,7 @@ class MyMostFrequentTaskFilter(Filter):
         for today's usage.
 
         :param QuerySet qs: The queryset to filter
-        :param int   value: number of frequest items
+        :param int   value: number of most frequent items
         :return:            The filtered queryset
         :rtype:             QuerySet
         """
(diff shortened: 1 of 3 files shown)


sliverc:
@kaldras 
On second thought I think having an ordering field frecency and an additional user filter would be the better solution instead of a new rather awkward filter. Pagination could then be used to limit result.

However this is way harder to implement and I think we need to move on. Therefore simply added a TODO so we won't forget... but let me know what you think.

timed/projects/filters.py
```diff
@@ -25,11 +27,44 @@ class Meta:
...
+
+    def filter(self, qs, value):
+        """Filter for given most frequently used tasks.
+
+        Most frequently used tasks are only counted within last
+        few months as older tasks are not relevant anymore
+        for today's usage.
+
+        :param QuerySet qs: The queryset to filter
+        :param int   value: number of frequest items
```
czosel: frequest = frequent request? :smile:

timed/projects/filters.py
```diff
@@ -25,11 +27,44 @@ class Meta:
...
+        :return:            The filtered queryset
+        :rtype:             QuerySet
+        """
+        user = self.parent.request.user
+        from_date = date.today() - timedelta(days=60)
+
+        qs = qs.filter(reports__user=user, reports__date__gt=from_date)
+        qs = qs.annotate(frequency=Count('reports')).order_by('-frequency')
+        qs = qs[:int(value)]
```
czosel: What is this line doing?

sliverc:
@czosel 
Adjusted as requested.

PR #25, opened 2017-07-18
Added additional filters to report end point

sliverc:
Broken backwards compatibility as reports are now not per default filtered
per user anymore. A user filter need to set manually to gain the same
behavior.

### DIFF
--- a/timed/tracking/permissions.py
+++ b/timed/tracking/permissions.py
@@ -1,0 +1,11 @@
+from rest_framework.permissions import SAFE_METHODS, BasePermission
+
+
+class IsOwnerOrReadOnly(BasePermission):
+    """Changing an object is only allowed if object belongs to current user."""
+
+    def has_object_permission(self, request, view, obj):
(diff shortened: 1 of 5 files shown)


sliverc:
@czosel Security issue fixed that all users can edit all reports... Should be mergable now.

timed/tracking/permissions.py
```diff
@@ -0,0 +1,14 @@
...
+    """Changing an object is only allowed if object belongs to current user."""
+
+    def has_object_permission(self, request, view, obj):
+        if request.method in SAFE_METHODS:
+            return True
+
+        if obj.user_id == request.user.id:
+            return True
+
+        return False
```
czosel: nitpick: `return obj.user_id == request.user.id`

sliverc: I could do a one liner as well.. but I left this out for now... :) Adjusted.

sliverc:
@czosel ready for review again

PR #26, opened 2017-07-19
Turning off forced requirement of docstrings

sliverc:
This led to a lot of noise in the code without adding any value. Whether a comment is required needs to be reviewed in a PR on a case by case basis and cannot be done with strict rules.


czosel:
:+1:

PR #27, opened 2017-07-21
Add export functionality to reports endpoint

sliverc:
Additional endpoint supporting exporting of reports into csv, ods and xlsx.

### DIFF
--- a/timed/tracking/tests/test_report.py
+++ b/timed/tracking/tests/test_report.py
@@ -2,15 +2,22 @@
 
 from datetime import date, timedelta
 
+import pyexcel
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 from django.utils.duration import duration_string
+from hypothesis import given, settings
+from hypothesis.extra.django import TestCase
+from hypothesis.extra.django.models import models
+from hypothesis.strategies import (builds, characters, dates, sampled_from,
+                                   timedeltas)
 from rest_framework.status import (HTTP_200_OK, HTTP_201_CREATED,
-                                   HTTP_204_NO_CONTENT, HTTP_401_UNAUTHORIZED,
-                                   HTTP_403_FORBIDDEN)
+                                   HTTP_204_NO_CONTENT, HTTP_400_BAD_REQUEST,
+                                   HTTP_401_UNAUTHORIZED, HTTP_403_FORBIDDEN)
 
-from timed.employment.factories import AbsenceTypeFactory, EmploymentFactory
-from timed.jsonapi_test_case import JSONAPITestCase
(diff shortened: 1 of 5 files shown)


timed/tracking/tests/test_report.py
```diff
@@ -242,3 +259,41 @@ def test_absence_update_on_create_report(self):
             Absence.objects.get(pk=absence.pk).duration ==
             employment.worktime_per_day - timedelta(hours=1)
         )
+
+
+class TestReportHypo(TestCase):
```
czosel: :+1: for using hypothesis

timed/tracking/tests/test_report.py
```diff
@@ -242,3 +259,41 @@ def test_absence_update_on_create_report(self):
...
+                min_delta=timedelta(0),
+                max_delta=timedelta(days=1)
+            )
+        )
+    )
+    @settings(timeout=5)
+    def test_report_export(self, file_type, report):
+        User.objects.create_user(username='test', password='1234qwer')
+        client = JSONAPIClient()
+        client.login('test', '1234qwer')
```
czosel: Are you creating a new user just to make sure that you see reports of other users?

timed/tracking/views.py
```diff
@@ -75,6 +78,38 @@ class ReportViewSet(ModelViewSet):
...
+                report.task.project.customer.name,
+                report.task.project.name,
+                report.task.name,
+                report.user.username,
+                report.comment,
+            ]
+            for report in queryset
+        ]
+
+        file_type = request.query_params.get('file_type')
```
czosel: Is it really fine to not check that `file_type` is in `['csv', 'xlsx', 'ods']`?

czosel:
And another nitpick: I always prefer to write commit messages in present tense, it makes the history much more readable (e.g. "add export functionality")

timed/tracking/tests/test_report.py
sliverc: Hypothesis doesn't support setUp methods properly hence I moved it into its own class which also needs to be subclassed of a special hypothesis TestCase to support transaction per example.

Hence it is needed to create a user here to be able to login.

czosel: Alright, thanks!

timed/tracking/views.py
sliverc: I agree. Adjusted.

sliverc:
Also added blacklist of Cs category in hypothesis test.

Read for review again.

czosel:
Awesome, LGTM :+1:

sliverc:
Merging then.

PR #28, opened 2017-07-24
Remove jet and crispy forms dependency

sliverc:
Jet and crispy forms do not add any more functionality but add higher complexity which makes updating of Django more difficult.


czosel:
:+1:

PR #30, opened 2017-07-24
For re-usability of timed modules defining deps in setuptools

sliverc:
This gives the ability to define timed based projects with additional features without redefining all dependencies again.


sliverc:
Not ready yet after all

sliverc:
@czosel 
Now it is ready...

PR #33, opened 2017-07-27
Add Timed specific User model

sliverc:
It is best practice to implement a project specific user model even when it is not used. This makes extending user easier in the future without an hassle of migration data.

This change breaks with migrations backwards comparability but should not be an issue as it is not in production yet.


sliverc:
should only be reviewed/merged once https://github.com/adfinis-sygroup/timed-backend/pull/32 is merged.

timed/employment/factories.py
```diff
@@ -33,13 +33,14 @@ def username(self):
...
-        model = settings.AUTH_USER_MODEL
+        model = get_user_model()
 
 
 class LocationFactory(DjangoModelFactory):
     """Location factory."""
 
-    name = Faker('city')
+    name = Faker('uuid4')
+    # cannot use city provider as name needs to be unique
```
czosel: :+1: i also saw this issue a couple of times

PR #34, opened 2017-07-27
Add possibility to configure supervisors per user

sliverc:
This is a first step to implement mail reports which will be sent out to supervisors e.g. when a employee has short time.


sliverc:
tests are added in #35 to gain coverage... closing this PR therefore and handling both in one.

PR #35, opened 2017-07-28
Add management command to notify supervisor on shorttime of their supervisees

sliverc:
This command is supposed to run regularly in a crontab.


timed/reporting/management/commands/notify_supervisors.py
```diff
@@ -0,0 +1,126 @@
...
+
+class Command(BaseCommand):
+    """
+    Notify supervisors command.
+
+    This command provides several methods usually used in a cron job
+    to notify supervisors on tracking status of their supervisees.
+    """
+
+    help = 'Notify supervisors on tracking status of user'
```
czosel: What is this for?

timed/reporting/management/commands/notify_supervisors.py
```diff
@@ -0,0 +1,126 @@
...
+        """Call given notification method and pass on options."""
+        method = getattr(self, options['notification'])
+        method(*options['options'])
+
+    def _decimal_hours(self, duration):
+        return duration.total_seconds() / 3600
+
+    def _get_supervisees_with_shorttime(self, start, end, ratio):
+        """
+        Get supervisees which reported less hours than they have should.
```
czosel: `should have`

timed/reporting/management/commands/notify_supervisors.py
```diff
@@ -0,0 +1,126 @@
...
+
+        :return: dict mapping all supervisees with shorttime with tuple of
+                 reported, expected, balance and actual ratio.
+        """
+        supervisees_shorttime = {}
+        supervisees = get_user_model().objects.all_supervisees()
+        for supervisee in supervisees:
+            worktime = self.serializer.get_worktime(supervisee, start, end)
+            reported, expected, balance = worktime
```
czosel: I'd drop `worktime` and make this a one-liner

timed/reporting/management/commands/notify_supervisors.py
sliverc: this help will be shown when running ./manage.py notify_supervisors --help

timed/reporting/management/commands/notify_supervisors.py
sliverc: this doesn't fit into 80 characters and I would have to add a new line at the dot... not really readable hence this solution.

timed/reporting/management/commands/notify_supervisors.py
sliverc: adjusted.

sliverc:
@czosel 
I have thought about the generic notify supervisors solution again and decided that it would be better to have a command for each notification. Reason being is that command line arguments are much better documented and parsed this way.
Hence I refactored the code a little bit to reflect this change. Best have a look again before merging.

PR #39, opened 2017-08-08
Configure django app with environment variables

sliverc:
First step to containerization. Per default for security reasons it runs in production environment which forces configuration of some settings. For development .env file with ENV=dev needs to be created. Such settings will use containers run by docker-compose.


sliverc:
@czosel Fixed CI and added possibility to configure ADMINS per environ as well. One test is still failing which is not related and is fixed in #40 Once that PR is merged and this updated CI should run.

czosel:
@sliverc Looks like there is still some coverage lacking...

sliverc:
@czosel Coverage fixed now.

PR #41, opened 2017-08-08
Check that migrations are created if necessary

sliverc:
No more missing migrations!


czosel:
Is this relevant when you're developing purely test-driven and make changes to a model?

sliverc:
Yep it unfortunately is... a use case can happen that missing migrations will be added in 2 different PRs which both have done models changes. This will lead to migrations conflict which can only be solved manually. This happens more than you think... :)

czosel:
I understand - thanks for the clarification!

sliverc:
@czosel finally CI running again.

PR #45, opened 2017-08-10
Fix the calculation of absence credits and balances.

anehx:


timed/employment/migrations/0005_merge_20170810_0924.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('employment', '0003_auto_20170808_1403'),
+        ('employment', '0004_auto_20170809_1647'),
+    ]
+
+    operations = [
+    ]
```
sliverc: I think we should try to keep the migrations clean. There seems to have been a conflict and I think it is better instead of a merging migration to delete the 3 migrations in this PR (0003, 0004 and 0005) and run makemigrations again.

timed/employment/serializers.py
```diff
@@ -5,7 +5,8 @@
 from dateutil import rrule
 from django.contrib.auth import get_user_model
 from django.utils.duration import duration_string
-from rest_framework_json_api.relations import ResourceRelatedField
+from rest_framework_json_api.relations import (ResourceRelatedField,  # noqa
+                                               SerializerMethodResourceRelatedField) # noqa
```
sliverc: Instead of the # noqa wouldn't it be better to import relations only?
`from from rest_framework_json_api import relations`

timed/employment/serializers.py
```diff
@@ -214,88 +228,176 @@ class Meta:
...
+    balance       = SerializerMethodField()
+
+    absence_credits = SerializerMethodResourceRelatedField(
+        source='get_absence_credits',
+        model=models.AbsenceCredit,
+        many=True,
+        read_only=True
+    )
+
+    user = SerializerMethodResourceRelatedField(
```
sliverc: I think it would be better to have a ForeignKey on the UserAbsenceType proxy model then this could SerializerMethodResourceRelatedField would be obsolete.

timed/employment/serializers.py
```diff
@@ -214,88 +228,176 @@ class Meta:
...
+        """Calculate the total used duration.
+
+        This is only calculated for types which fill the worktime such as
+        sickness.
+
+        :return: The total used duration
         :rtype:  str
         """
-        used = self.get_used_raw(instance)
+        if not instance.fill_worktime:
```
sliverc: Maybe a comment refering to get_used_duration might be helpful to better understand this line.

timed/employment/serializers.py
```diff
@@ -214,88 +228,176 @@ class Meta:
...
+            return None
 
-        return duration_string(used)
+        start, end = self.get_dates(instance)
 
-    def get_balance(self, instance):
-        """Format the balance.
+        return duration_string(sum(Absence.objects.filter(
```
sliverc: Calculation of the Sum should be done on the database. I guess easiest is to extend the UserAbsenceType proxy model and do another annotate with Sum. start and end date need to be passed on to with_user

timed/employment/serializers.py
```diff
@@ -214,88 +228,176 @@ class Meta:
...
+    def get_balance(self, instance):
+        """Calculate the balance of credited and used days.
 
-        :return: The formatted balance
-        :rtype:  str or None
+        :return: The balance
+        :rtype:  int
         """
-        balance = self.get_balance_raw(instance)
+        credit = self.get_credit(instance)
```
sliverc: Currently credit needs to be calculated twice. Once for the serializer and once to calculate the balance.
When calculation of sum will be done on database as described above the credit would only be calculated once.

timed/employment/tests/test_user.py
```diff
@@ -213,3 +214,61 @@ def test_user_without_employment(self):
...
         )
+
+    def test_user_absence_types(self):
+        credits = AbsenceCreditFactory.create_batch(3, user=self.user)
+
+        url = reverse('user-detail', args=[
+            self.user.id
+        ])
+
+        res = self.client.get('{0}?include=user_absence_types'.format(url))
```
sliverc: get params should be passed on as data to .get instead of string formatting.

timed/employment/tests/test_user.py
```diff
@@ -213,3 +214,61 @@ def test_user_without_employment(self):
...
+
+        assert (
+            inc[0]['id'] ==
+            '{0}-{1}'.format(self.user.id, credits[0].absence_type.id)
+        )
+
+        assert type(inc[0]['attributes']['credit']) == int
+        assert type(inc[0]['attributes']['used-days']) == int
+        assert type(inc[0]['attributes']['balance']) == int
+        assert inc[0]['attributes']['used-duration'] is None
```
sliverc: We need to test whether the attributes have been calculated correctly. Instead of random data some fixed data would make this easily possible.

timed/employment/tests/test_user.py
```diff
@@ -213,3 +214,61 @@ def test_user_without_employment(self):
...
+
+        assert (
+            inc[0]['id'] ==
+            '{0}-{1}'.format(self.user.id, absence_type.id)
+        )
+
+        assert inc[0]['attributes']['credit'] is None
+        assert inc[0]['attributes']['used-days'] is None
+        assert inc[0]['attributes']['balance'] is None
+        assert type(inc[0]['attributes']['used-duration']) == str
```
sliverc: Also here calculation of used-duration needs to be verified.

timed/employment/serializers.py
```diff
@@ -16,9 +17,51 @@
...
+
+        end = datetime.strptime(
+            request.query_params.get(
+                'until',
+                date.today().strftime('%Y-%m-%d')
+            ),
+            '%Y-%m-%d'
+        ).date()
+
+        user = get_user_model().objects.get(pk=instance.id)
```
sliverc: I am confused... Instance is a user already right? So this call is not needed.

timed/employment/tests/test_user.py
```diff
@@ -213,3 +214,95 @@ def test_user_without_employment(self):
...
+
+        AbsenceFactory.create(date=date.today() - timedelta(days=1),
+                              user=self.user,
+                              type=absence_type)
+
+        url = reverse('user-detail', args=[
+            self.user.id
+        ])
+
+        res = self.client.get('{0}?include=user_absence_types'.format(url))
```
sliverc: get params should be passed on as data to .get instead of string formatting.

timed/employment/models.py
```diff
@@ -179,3 +181,91 @@ class OvertimeCredit(models.Model):
...
     date     = models.DateField()
     duration = models.DurationField(blank=True, null=True)
+
+
+class UserAbsenceTypeManager(models.Manager):
+    def with_user(self, user, start_date, end_date):
+        from timed.tracking.models import Absence
+
+        return UserAbsenceType.objects.raw("""
+            SELECT
```
sliverc: I like it this way now... maybe a comment would be good explaining why using raw query. And I think sq1 and sq2 should be renamed for better readability of the SQL (the other aliases are a bit more obvious what they mean).
Otherwise looks great.

PR #46, opened 2017-08-10
JWT needs to be authenticated before django admin session

sliverc:
Otherwise when exporting a report it would take authentication of session which is not desired.


czosel:
Out of curiosity: Why is that?

sliverc:
When you are logged in as Django Admin there is a cookie set. When now a file is downloaded with a jwt it will try to authenticated in the order how the authentication classes are defined - so it will take the Django Admin cookie. In my case I was logged in as Admin and this user was then set as creator of a work report instead of user which was logged in really.

czosel:
Ah, that makes sense. Thanks!

PR #49, opened 2017-08-15
Add verification logic to report end point

sliverc:
This includes:
* verified_by field on report
* filters for not_verified on reports endpoint
* action to verify list or detail of reports


sliverc:
WIP for now. Need to add missing test first for filter not_verified.

sliverc:
@anehx Test added so ready for review.

sliverc:
@anehx There was an error with verifying page after all which is now fixed

PR #59, opened 2017-08-18
Use number filter for booleans

anehx:


sliverc:
@anehx Do not forget the tests... :)

PR #87, opened 2017-09-13
Convert to and from datetime on ActivityBlock to time fields

sliverc:
An activity is on a particular day so a block may only be added within this day.


sliverc:
@anehx As this breaks backwards compatibility in terms of frontend this should only be merged once frontend has been adjusted.

anehx:
Also, if the frontend needs to split overlapping activities, the `date` of the activity can't be set automatically..

sliverc:
@anehx 
I have fixed issue with migration of blocks where `to_datetime` is None.
About your second comment: I am not quite sure what you mean by being set automatically. API expects date to be assigned by caller. Maybe I misunderstand your comment though.

sliverc:
@anehx 
OK as discussed I have removed the auto_add on activity date.

anehx:
We need to wait with merging this until the [frontend PR](https://github.com/adfinis-sygroup/timed-frontend/pull/115) is merged.

PR #98, opened 2017-10-09
Add filters and fields for the user overview

anehx:
This PR is needed for frontend [PR #116](https://github.com/adfinis-sygroup/timed-frontend/pull/116) to work

### DIFF
--- a/timed/employment/filters.py
+++ b/timed/employment/filters.py
@@ -1,7 +1,7 @@
 """Filters for filtering the data of the employment app endpoints."""
 
 
-from django_filters import DateFilter, Filter, FilterSet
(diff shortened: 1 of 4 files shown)


timed/employment/filters.py
```diff
@@ -37,6 +37,9 @@ class Meta:
 
 
 class UserFilterSet(FilterSet):
+    active = NumberFilter(name='is_active')
+    supervisor = Filter(name='supervisors__id', lookup_expr='contains')
```
sliverc: This is a fairly fancy filter I think it would be good to add a test (best in pytest style)

anehx: Done!

PR #103, opened 2017-10-12
Integrate current balance and balance in shorttime report

sliverc:


timed/reports/management/commands/notify_supervisors_shorttime.py
```diff
@@ -90,7 +94,10 @@ def _get_supervisees_with_shorttime(self, start, end, ratio):
                     self._decimal_hours(reported),
                     self._decimal_hours(expected),
                     self._decimal_hours(balance),
-                    supervisee_ratio
+                    supervisee_ratio,
+                    self._decimal_hours(
+                        supervisee.calculate_worktime(start_year, today)[2]
+                    ),
```
winged: Depending on when the cronjob is run (morning, evening), this will yield different results.. I would prefer the balance of "yesterday" as it's more likely to be correct. If it's run in the morning, you will get a balance 8.5h below the actual balance, for example

timed/reports/management/commands/notify_supervisors_shorttime.py
```diff
@@ -74,10 +74,14 @@ def _get_supervisees_with_shorttime(self, start, end, ratio):
         Get supervisees which reported less hours than they should have.
 
         :return: dict mapping all supervisees with shorttime with tuple of
-                 reported, expected, balance and actual ratio.
+                 reported, expected, balance, actual ratio and current balance.
```
winged: I think a nested dict would be more useful than a tuple here.

Also note the docstring of _notify_supervisors() further below does not match this list

timed/reports/management/commands/notify_supervisors_shorttime.py
winged: Oh and is worktime since the start of the year always correct? Shouldn't we take the employment into consideration as well?

timed/reports/templates/mail/notify_supervisor_shorttime.txt
```diff
@@ -3,5 +3,5 @@ Time range: {{start}} - {{end}}
 Ratio: {{ratio}}
 
 {% for suspect, worktime in suspects %}
-{{suspect.get_full_name}} {{worktime.0}}/{{worktime.1}} (Ratio {{worktime.3|floatformat:2}} Balance {{worktime.2}})
+{{suspect.get_full_name}} {{worktime.0}}/{{worktime.1}} (Ratio {{worktime.3|floatformat:2}} Range Balance {{worktime.2}} Balance {{worktime.4}})
```
winged: Q: What is "range balance"?

sliverc: The balance of the given time range... have a better idea for a name?

timed/reports/management/commands/notify_supervisors_shorttime.py
sliverc: calculate_worktime already takes care of the employment length...  I will change end date to yesterday.

sliverc: Thought about it again. notify supervisor has an offset parameter which is set because it is expected that users may report their time too late. This needs to be taken into consideration as well for current balance. I changed the code accordingly.

timed/reports/templates/mail/notify_supervisor_shorttime.txt
winged: hours worked? If that is what it does

timed/reports/management/commands/notify_supervisors_shorttime.py
sliverc: there is no built in way to access dicts in django templates - that's why I have chosen to use a tuple.

timed/reports/management/commands/notify_supervisors_shorttime.py
winged: Maybe even better: calculate the balance at the end of the analyzed time period

timed/reports/templates/mail/notify_supervisor_shorttime.txt
sliverc: no this is the balance between expected - reported in given time range.

timed/reports/management/commands/notify_supervisors_shorttime.py
sliverc: that's how I have done it now.

timed/reports/management/commands/notify_supervisors_shorttime.py
winged: You can access dict keys just like object properties

timed/reports/templates/mail/notify_supervisor_shorttime.txt
winged: so, a delta then?

sliverc: Delta sounds good. Changed.

timed/reports/management/commands/notify_supervisors_shorttime.py
sliverc: Changed.

sliverc:
@winged All changes should be done now and can be reviewed again.

PR #104, opened 2017-10-12
Add option to configure server email address

sliverc:


timed/settings.py
```diff
@@ -242,6 +242,11 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
     default('webmaster@localhost')
 )
 
+SERVER_EMAIL = env.str(
+    'DJANGO_SERVER_EMAIL',
+    default('root@localhost')
+)
+
```
winged: Side note: It would probably be a good idea to have the configurable values documented somewhere

sliverc: True... I put it on my list.

PR #105, opened 2017-10-12
Integrate redmine and subscription apps

sliverc:
Integrated separately developed redmine and subscription module. Those two modules we tightly coupled so integrated them together in one PR.

Goal is to eliminate private timed apps main reasons to simplify containerization and development in the future.

Only code which actually changed is in commit 8c7371e


winged:
Isn't this whole thing supposed to be in our Gitlab?

sliverc:
this caused several issues with deployment and this way we have to duplicate docker container code, helm scripts etc.
there is nothing confidential in here so even though at first I thought it is a good idea to keep it internal but it causes more work than it helps. And the code might be useful for others as well.

PR #106, opened 2017-10-13
WIP: Add subscription api end points

sliverc:
In this step merging SubscriptionProject and BillingType for clean database scheme.


sliverc:
see #114

PR #114, opened 2017-10-18
Add subscription api end points

sliverc:
In this step merging SubscriptionProject and BillingType


sliverc:
@anehx @velrest 
Rebased and fixed an issue that during migration all customer passwords wouldn't have been deleted...

Otherwise all other changes look good to me. If one or both of you could review this that would be great. And then you can merge whenever you are ready for deployment.

PR #116, opened 2017-10-23
Move worktime calculation into its own end point

sliverc:
Give client the ability to calculate worktime on a given date. This PR includes some minor performance improvements in the calculation as well.

User end point won't include a worktime balance anymore. This means this change is not backwards compatible and should not be merged before Frontend has not been updated.

### DIFF
--- a/timed/mixins.py
+++ b/timed/mixins.py
@@ -1,0 +1,72 @@
+from rest_framework_json_api.relations import ResourceRelatedField
+
+
+class AggregateQuerysetMixin(object):
+    """
+    Add support for aggregate queryset in view.
+
+    Wrap queryst dicts into aggregate object to support renderer
+    which expect attributes.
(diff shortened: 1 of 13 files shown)


timed/mixins.py
```diff
@@ -0,0 +1,72 @@
...
+class AggregateQuerysetMixin(object):
+    """
+    Add support for aggregate queryset in view.
+
+    Wrap queryst dicts into aggregate object to support renderer
+    which expect attributes.
+    It additional prefetches related instances represented as id in
+    aggregate.
+
+    In aggregates only an id of a related field is part of the object.
```
anehx: "In" => "It"?

sliverc: Well spotted... :) I rewrote the doc string a little bit to make it more comprehensible.

PR #119, opened 2017-10-25
Move absence balances into its own view

sliverc:
This gives the possibility to retrieve absence balances from a given date.


sliverc:
@anehx Added protection level so only self, superuser and supervisors can see absence balances. Should be ready now... let me know if anything is missing.

PR #122, opened 2017-10-30
Add last reported date filter to worktime balance

sliverc:
Only returns worktime balances of every user at the date when either a report or a absence has been reported.

### DIFF
--- a/timed/employment/serializers.py
+++ b/timed/employment/serializers.py
@@ -3,18 +3,18 @@
 from datetime import date, timedelta
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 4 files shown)


timed/employment/serializers.py
```diff
@@ -67,17 +67,40 @@ class Meta:
 
 
 class WorktimeBalanceSerializer(Serializer):
-    date    = serializers.DateField()
+    date = SerializerMethodField()
+    # date    = serializers.DateField()
```
anehx: Please remove line 71 here..

PR #136, opened 2017-11-09
Add report-intersection end point

sliverc:
Returns a report-intersection resource with fields which are equal in result. If result has different values of one field None will be returned.

### DIFF
--- a/timed/mixins.py
+++ b/timed/mixins.py
@@ -1,5 +1,7 @@
 from rest_framework_json_api import relations
 
+from timed.serializers import AggregateObject
+
 
 class AggregateQuerysetMixin(object):
     """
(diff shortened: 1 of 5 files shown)


timed/mixins.py
```diff
@@ -70,7 +60,7 @@ def get_serializer(self, data, *args, **kwargs):
 
         # enhance entry dicts with model instances
         data = [
-            self.AggregateObject(**{
+            AggregateObject(**{
```
winged: That double-dicting is not really necessary. You can get rid of the "outer" dict, it will work the same.

timed/tracking/serializers.py
```diff
@@ -188,6 +190,87 @@ class Meta:
         ]
 
 
+class ReportIntersectionSerializer(Serializer):
+    """Serializers intersection of given reports."""
```
winged: I think this needs a bit more explanation. What exactly does this *do*? What *is* a report intersection?

sliverc: I added some more documentation which I hope makes things clearer. It should get even more clear in future PR where we will introduce bulk updates on reports.

The intersection is mainly needed for frontend to tell user which fields are already the same in given reports before bulk updating.

timed/mixins.py
sliverc: I do not see how this should work in this particular case... because what is being done here is to combine two dicts into one before passing it to AggregateObject.

But maybe you meant creation of AggregateObject in tracking/views.py... I removed the outer dict there which was definitely not necessary.

winged: No I just meant that pattern -  `**{**{}, **{}}`.

PR #144, opened 2017-11-13
Add id list and verifier filter

sliverc:


timed/filters.py
```diff
@@ -0,0 +1,13 @@
+from django_filters import Filter
+
+
+class ListFilter(Filter):
```
winged: I would rename this to something more self-explanatory.. like `CommaSeparatedListFilter`. I know it's much longer, but `ListFilter` appears too generic in my opinion.

sliverc: When looking for a better name I noticed that there is already a filter called `BaseInFilter` in django filters which does what `ListFIlter` does. So removed `ListFilter`.

PR #145, opened 2017-11-13
Enforce that only superuser may change verified reports

sliverc:
This protects changing of reports which may already be billed.

### DIFF
--- a/timed/tracking/filters.py
+++ b/timed/tracking/filters.py
@@ -122,9 +122,11 @@ class ReportFilterSet(FilterSet):
 
             # only owner, reviewer or supervisor may change reports
             queryset = queryset.filter(
-                Q(user__supervisors=user) |
-                Q(task__project__reviewers=user) |
-                Q(user=user)
+                (
+                    Q(user__supervisors=user) |
+                    Q(task__project__reviewers=user) |
+                    Q(user=user)
+                ) & Q(verified_by__isnull=True)
             )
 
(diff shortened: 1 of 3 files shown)


timed/tracking/filters.py
```diff
@@ -122,9 +122,11 @@ def filter_editable(self, queryset, name, value):
 
             # only owner, reviewer or supervisor may change reports
```
winged: Maybe should update the comment as well to reflect the implemented policy

timed/tracking/filters.py
```diff
@@ -134,9 +136,11 @@ def filter_editable(self, queryset, name, value):
...
 
             queryset = queryset.exclude(
-                Q(user__supervisors=user) |
-                Q(task__project__reviewers=user) |
-                Q(user=user)
+                (
+                    Q(user__supervisors=user) |
+                    Q(task__project__reviewers=user) |
+                    Q(user=user)
+                ) & Q(verified_by__isnull=True)
```
winged: Can we get rid of this code duplication?

timed/tracking/filters.py
sliverc: adjusted.

timed/tracking/filters.py
sliverc: Agree. Didn't like this before either. Moved filter query into its own method.

PR #179, opened 2017-11-28
Add test to authenticate with ldap backend

sliverc:


timed/conftest.py
```diff
@@ -1,9 +1,41 @@
+import mockldap
```
winged: Nice, didn't know about that one! Good find :)

pytest.ini
```diff
@@ -1,3 +1,7 @@
 [pytest]
 DJANGO_SETTINGS_MODULE=timed.settings
 addopts = --reuse-db
+env =
+  DJANGO_AUTH_LDAP_ENABLED=True
+  DJANGO_AUTH_LDAP_SERVER_URI=ldap://127.0.0.1
+  DJANGO_AUTH_LDAP_USER_DN_TEMPLATE=uid=%(user)s,ou=people,o=test
```
winged: If we're enabling LDAP auth here and then testing that, are we even testing the regular DB auth?

timed/conftest.py
```diff
@@ -1,9 +1,41 @@
...
+    )
+
+    directory = dict([top, people, groups, ldapuser])
+    mock = mockldap.MockLdap(directory)
+    mock.start()
+
+    yield
+
+    mock.stop()
+    del mock
```
winged: `del mock` shouldn't be necessary, the scope ends here anyway

pytest.ini
sliverc: Yes... because it enables two backends. First it will always tries ldap backend if user doesn't exist it will try db backend. The same way as it also runs on Timed Production Server.

timed/conftest.py
sliverc: Was part of the example but true... removed it.

PR #183, opened 2017-12-06
Add reviewers to project serializer

anehx:


anehx:
@sliverc I'm kind of in over my head with those failing tests.. Could you take a look at it? I need that includes to determine whether a report is editable or not in the frontend.

PR #200, opened 2018-01-04
WIP: Adjust hiding of overlapping public holidays and absences 

sliverc:
Determine by location of employment at the date of absence whether a public holiday overlaps it.

For this fix to work following Django bug needs to be fixed first:
https://code.djangoproject.com/ticket/28442

As this fix will most likely only be available in Django 2.0, started to fix deprecation warning so Timed is compatible.

Work in progress.


velrest:
@trowik Is this still an issue or has this been implemented already?

trowik:
@velrest this hasn't been implemented

PR #275, opened 2018-06-21
Bump to version 0.10.0

velrest:


timed/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '0.9.3'
+__version__ = '0.10.0
```
sliverc: a quote is missing here...

PR #280, opened 2018-06-29
fix package filter by replacing a JOIN with a subquery

velrest:
previously the filter returned too many packages because of a JOIN

### DIFF
--- a/timed/subscription/filters.py
+++ b/timed/subscription/filters.py
@@ -1,10 +1,17 @@
 from django_filters import FilterSet, NumberFilter
 
+from timed.projects.models import Project
+
 from . import models
 
 
 class PackageFilter(FilterSet):
-    customer = NumberFilter(field_name='billing_type__projects__customer')
+    customer = NumberFilter(method='filter_customer')
+
+    def filter_customer(self, queryset, name, value):
+        billing_types = (Project.objects.filter(customer=value)
+                         .values('billing_type'))
+        return queryset.filter(billing_type__in=billing_types)
 
     class Meta:
         model = models.Package
--- a/timed/subscription/filters.py
+++ b/timed/subscription/filters.py
@@ -9,8 +9,11 @@ class PackageFilter(FilterSet):
     customer = NumberFilter(method='filter_customer')
 
     def filter_customer(self, queryset, name, value):
-        billing_types = (Project.objects.filter(customer=value)
-                         .values('billing_type'))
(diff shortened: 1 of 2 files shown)


timed/subscription/tests/test_package.py
```diff
@@ -18,11 +19,15 @@ def test_subscription_package_list(auth_client):
...
 def test_subscription_package_filter_customer(auth_client):
-    other_project = ProjectFactory.create()
-    PackageFactory.create(billing_type=other_project.billing_type)
-
-    my_project = ProjectFactory.create()
-    package = PackageFactory.create(billing_type=my_project.billing_type)
+    customer = CustomerFactory.create()
+    billing_type = BillingTypeFactory.create()
+    package = PackageFactory.create(billing_type=billing_type)
+    ProjectFactory.create(billing_type=billing_type, customer=customer)
```
sliverc: you could use bulk_create and assign 2 so two are created.

timed/subscription/tests/test_package.py
```diff
@@ -18,11 +19,15 @@ def test_subscription_package_list(auth_client):
...
+
+    my_project = ProjectFactory.create(
```
sliverc: my_project var is not needed as below you could simply adjust it to user create customer

timed/subscription/filters.py
```diff
@@ -1,10 +1,17 @@
...
+
 from . import models
 
 
 class PackageFilter(FilterSet):
-    customer = NumberFilter(field_name='billing_type__projects__customer')
+    customer = NumberFilter(method='filter_customer')
+
+    def filter_customer(self, queryset, name, value):
+        billing_types = (Project.objects.filter(customer=value)
```
sliverc: what is the reason to use braces here?

sliverc:
Ahh end you did branch this of current master it seems... needs rebasing ;)

timed/subscription/tests/test_package.py
velrest: the second project is to check that only one package is returned.

timed/subscription/filters.py
velrest: the line is too long

sliverc: I guess this is matter of taste... but something like the following is properly more readable:
```python
        billing_types = Project.objects.filter(
            customer=value
        ).values(
            'billing_type'
        )
```

PR #281, opened 2018-06-29
add customer_visible flag to project

velrest:


timed/subscription/views.py
```diff
@@ -24,10 +23,8 @@ class SubscriptionProjectViewSet(viewsets.ReadOnlyModelViewSet):
     )
 
     def get_queryset(self):
-        queryset = Project.objects.annotate(
-            count_packages=Count('billing_type__packages')
-        )
-        return queryset.filter(archived=False, count_packages__gt=0)
+        return Project.objects.filter(
+            archived=False, customer_visible=True)
```
sliverc: a small nitpick: brace should be indented on the next line... it is more consistent to rest of module and in my opinion easier editable... ;).

timed/subscription/tests/test_subscription_project.py
```diff
@@ -25,7 +26,7 @@ def test_subscription_project_list(auth_client):
     # not billable reports should not be included in spent hours
     ReportFactory.create(not_billable=True, task=task,
                          duration=timedelta(hours=4))
-    # project of same customer but without a billing type with packages
+    # project of same customer but without customer_visible=True with packages
```
sliverc: sorry I did not see this before... but this is an odd comment :).

Maybe better `project of same customer but without customer_visible set`

PR #290, opened 2018-07-17
Add is_supervisor filter

Yelinz:
The filter return users who are supervisors or not.

### DIFF
--- a/timed/employment/filters.py
+++ b/timed/employment/filters.py
@@ -1,11 +1,12 @@
 from datetime import date
 
-from django.db.models import Value
+from django.db.models import Exists, OuterRef, Value
 from django.db.models.functions import Coalesce
 from django_filters.rest_framework import (DateFilter, Filter, FilterSet,
                                            NumberFilter)
 
 from timed.employment import models
+from timed.projects.models import Project
 
 
 class YearFilter(Filter):
@@ -49,12 +50,19 @@ class AbsenceTypeFilterSet(FilterSet):
 
 
 class UserFilterSet(FilterSet):
-    active = NumberFilter(field_name='is_active')
-    supervisor = NumberFilter(field_name='supervisors')
+    active      = NumberFilter(field_name='is_active')
+    supervisor  = NumberFilter(field_name='supervisors')
+    is_reviewer = NumberFilter(method='filter_reviewers')
+
+    def filter_reviewers(self, queryset, name, value):
+        reviewer = Project.objects.filter(reviewers=OuterRef('pk'))
+        return queryset.annotate(
+            is_reviewer=Exists(reviewer)
+        ).filter(is_reviewer=value)
 
     class Meta:
         model  = models.User
-        fields = ['active', 'supervisor']
+        fields = ['active', 'supervisor', 'is_reviewer']
 
 
 class EmploymentFilterSet(FilterSet):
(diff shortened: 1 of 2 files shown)


timed/employment/filters.py
```diff
@@ -53,13 +54,20 @@ class UserFilterSet(FilterSet):
...
     is_reviewer = NumberFilter(method='filter_reviewers')
+    is_supervisor = NumberFilter(method='filter_supervisors')
 
     def filter_reviewers(self, queryset, name, value):
         reviewer = Project.objects.filter(reviewers=OuterRef('pk'))
         return queryset.annotate(
             is_reviewer=Exists(reviewer)
         ).filter(is_reviewer=value)
 
+    def filter_supervisors(self, queryset, name, value):
```
sliverc: Didn't see this before but this comment also concerns filter_reviewers:
I think the function name should be called like the filter is called. Otherwise is there a confusion with `supervisor`.

In this example it would be `filter_is_supervisor`

timed/employment/filters.py
```diff
@@ -53,13 +54,20 @@ class UserFilterSet(FilterSet):
...
+        supervisor = User.objects.filter(supervisors=OuterRef('pk'))
```
sliverc: This does exactly do what you expect I think. It would return all users which have a supervisor.

I have just found that there are actually methods in UserManager for `all_reviewers` and `all_supervisors`.

This could then look something like this `queryset.filter(pk__in=User.objects.all_supervisors)`.

Yelinz:
The change, to using the UserManager, made that the filters only give back supervisors and reviewers. You cannot use them to get users who are the opposite, not supervisors and reviewers.

timed/employment/filters.py
```diff
@@ -53,24 +52,18 @@ class Meta:
 class UserFilterSet(FilterSet):
     active      = NumberFilter(field_name='is_active')
     supervisor  = NumberFilter(field_name='supervisors')
-    is_reviewer = NumberFilter(method='filter_reviewers')
-    is_supervisor = NumberFilter(method='filter_supervisors')
+    is_reviewer = NumberFilter(method='filter_is_reviewers')
```
sliverc: shouldn't this be `filter_is_reviewer` without a `s`? Same below.

timed/employment/filters.py
```diff
@@ -53,24 +52,18 @@ class Meta:
...
+    is_supervisor = NumberFilter(method='filter_is_supervisors')
 
-    def filter_reviewers(self, queryset, name, value):
-        reviewer = Project.objects.filter(reviewers=OuterRef('pk'))
-        return queryset.annotate(
-            is_reviewer=Exists(reviewer)
-        ).filter(is_reviewer=value)
+    def filter_is_reviewers(self, queryset, name, value):
+        return queryset.filter(pk__in=User.objects.all_reviewers())
```
sliverc: when value is 0 we should use exclude instead of filter to invert query. Same below.

timed/employment/filters.py
```diff
@@ -52,14 +52,20 @@ class Meta:
...
-
-    def filter_is_supervisors(self, queryset, name, value):
-        return queryset.filter(pk__in=User.objects.all_supervisors())
+    is_reviewer = NumberFilter(method='filter_is_reviewer')
+    is_supervisor = NumberFilter(method='filter_is_supervisor')
+
+    def filter_is_reviewer(self, queryset, name, value):
+        if value:
+            return queryset.filter(pk__in=User.objects.all_reviewers())
+        else:
```
sliverc: This looks functional good now. Just one small thing: I usually try to avoid `else` whenever possible not to complicate the code.

This else could actually be easily removed as the return statement above already exist the method. Same below.

sliverc:
Great, let' merge it.

PR #292, opened 2018-07-20
Update django-filter to 2.0.0

Yelinz:


timed/employment/filters.py
```diff
@@ -13,6 +14,8 @@ class YearFilter(Filter):
     """Filter to filter a queryset by year."""
 
     def filter(self, qs, value):
+        if value in EMPTY_VALUES:
```
sliverc: the doc string should still come right after the method definition.

Although in this case actually as the doc string doesn't say anything really you can also remove it.

timed/projects/filters.py
```diff
@@ -51,6 +52,8 @@ class MyMostFrequentTaskFilter(Filter):
     """
 
     def filter(self, qs, value):
+        if value in EMPTY_VALUES:
+            return qs
```
sliverc: same as above doc string needs to come right after function. However this actually has some information and needs to be kept.

timed/tracking/filters.py
```diff
@@ -18,13 +19,16 @@ def boolean_filter(func):
     """
     @wraps(func)
     def wrapper(self, qs, value):
+        if value in EMPTY_VALUES:
+            return qs
```
sliverc: same here doc string can be removed though as it is logical by the code.

PR #301, opened 2018-08-10
Migrate activity-block into activity

Yelinz:
Include activity-block into activity for the new frontend structure


timed/tracking/migrations/0007_migrate_activity_blocks.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations
+
+
+def migrate_blocks(apps, schema_editor):
+    Activity = apps.get_model('tracking', 'Activity')
+    for activity in Activity.objects.all():
+        if activity.blocks.count() == 1:
```
sliverc: I have a feeling this will lead to duplication of activities as the activity which has more then one block will stay behind.

I guess the first activity block is always an update and the others should be added.

timed/tracking/migrations/0007_migrate_activity_blocks.py
```diff
@@ -0,0 +1,30 @@
...
+            block = activity.blocks.all()[0]
+            activity.from_time = block.from_time
+            activity.to_time = block.to_time
+            activity.save()
+        else:
+            for block in activity.blocks.all():
+                Activity(from_time=block.from_time,
```
sliverc: you have to do Activity.objects.create() otherwise nothing is created.

timed/tracking/serializers.py
```diff
@@ -22,56 +21,30 @@
 class ActivitySerializer(ModelSerializer):
     """Activity serializer."""
 
-    duration = DurationField(read_only=True)
     user     = CurrentUserResourceRelatedField()
     task     = ResourceRelatedField(queryset=Task.objects.all(),
```
sliverc: as we are working on this code bit... this definition of task is actually not needed as this will be the default. This can be removed and serializer will do it automatically.

Yelinz: Done

timed/tracking/migrations/0007_migrate_activity_blocks.py
Yelinz: Changed it

timed/tracking/migrations/0007_migrate_activity_blocks.py
Yelinz: Changed it so that there is an index, so it updates on the first iteration and on more it creates.

sliverc:
This looks good. One thing maybe to consider in the future is to make transferred flag read only and move the transfer login to the backend.

PR #312, opened 2018-08-31
Add arguments cc and message

JohnL17:


timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -48,10 +48,24 @@ def add_arguments(self, parser):
...
         )
+        parser.add_argument(
+            '--message',
+            type=str,
+            dest='message',
+            help='Additional message to send if there are unverified reports'
+        )
+        parser.add_argument(
+            '--cc',
+            type=str,
```
sliverc: for completeness shoudln't this be a list of addresses?

timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -48,10 +48,24 @@ def add_arguments(self, parser):
...
+            dest='cc',
+            help='Email address where to send a cc'
+        )
 
     def handle(self, *args, **options):
         months = options['months']
         offset = options['offset']
+        message = options['message']
```
sliverc: not sure but what happens when I do not pass on `--message`. Will this be value then None which is then added to the email? I guess something to verify.

timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -95,13 +109,14 @@ def _notify_reviewers(self, start, end, reports):
...
                         'end': str(end),
+                        'message': message,
                         'reviewer': reviewer,
                         'protocol': settings.HOST_PROTOCOL,
                         'domain': settings.HOST_DOMAIN,
                     }, using='text'
                 )
 
-                mails.append((subject, body, from_email, [reviewer.email]))
+                mails.append((subject, body, from_email, [reviewer.email, cc]))
```
sliverc: this will be `To` in the mail but I guess this should be cc?

timed/reports/tests/test_notify_reviewers_unverified.py
```diff
@@ -27,14 +27,27 @@ def test_notify_reviewers(db, mailoutbox):
...
+            msg = item.strip()
+
+    assert len(mailoutbox) == 1
     assert mail.to == [reviewer_work.email]
     url = (
         'http://localhost:4200/analysis?fromDate=2017-07-01&'
         'toDate=2017-07-31&reviewer=%d&editable=1'
     ) % reviewer_work.id
     assert url in mail.body
+    assert msg == 'This is a test'
```
sliverc: wouldn't be easier to test:

```python
assert 'This is a test' in mail.body
```

timed/reports/management/commands/notify_reviewers_unverified.py
JohnL17: I have set the default of message to an empty string.

timed/reports/management/commands/notify_reviewers_unverified.py
JohnL17: I changed it as discussed.

timed/reports/tests/test_notify_reviewers_unverified.py
JohnL17: Done.

JohnL17:
@sliverc I'm on it

timed/reports/tests/test_notify_reviewers_unverified.py
```diff
@@ -27,14 +27,27 @@ def test_notify_reviewers(db, mailoutbox):
     ReportFactory.create(date=date(2017, 7, 1), task=task_no_work,
                          verified_by=reviewer_no_work)
 
-    call_command('notify_reviewers_unverified')
+    call_command(
```
winged: Please add tests for all call variants (with/without cc, with/without message)

JohnL17: Done

JohnL17:
Ready for review again.

timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -95,13 +112,25 @@ def _notify_reviewers(self, start, end, reports):
...
-                mails.append((subject, body, from_email, [reviewer.email]))
+                message = EmailMessage(
+                    subject=subject,
+                    body=body,
+                    from_email=from_email,
+                    to=[reviewer.email],
+                    connection=connection
+                )
+
+                if cc:
```
sliverc: this `if` is not necessary. Just make no default for cc as EmailMessage expects None to cc for not configured.

timed/reports/tests/test_notify_reviewers_unverified.py
```diff
@@ -27,14 +27,66 @@ def test_notify_reviewers(db, mailoutbox):
...
+    ) % reviewer_work.id
+    assert url in mail.body
+    assert 'This is a test' in mail.body
+    assert mail.cc == ['example@example.com']
+
+    call_command('notify_reviewers_unverified',)
+
+    # checks
+    mail = mailoutbox[1]
+    assert len(mailoutbox) == 2
```
sliverc: best to use one test method for each test. Or even better use `pytest.parametrize` when possible.

timed/reports/tests/test_notify_reviewers_unverified.py
```diff
@@ -27,14 +27,66 @@ def test_notify_reviewers(db, mailoutbox):
...
+    call_command(
+        'notify_reviewers_unverified',
+        '--cc=example@example.com',
+        '--message=This is a test'
+    )
 
     # checks
-    assert len(mailoutbox) == 1
     mail = mailoutbox[0]
+    assert len(mailoutbox) == 1
```
sliverc: I think this move happen by mistake right? assert len check should be before accessing mailoutbox otherwise we end up with a invalid index error.

timed/reports/management/commands/notify_reviewers_unverified.py
JohnL17: Done.

timed/reports/tests/test_notify_reviewers_unverified.py
JohnL17: I have changed that.

timed/reports/tests/test_notify_reviewers_unverified.py
JohnL17: Yes this was a mistake. I have corrected it already.

JohnL17:
Ready for a new review

timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -80,13 +96,17 @@ def _get_unverified_reports(self, start, end):
...
+    def _notify_reviewers(self, start, end, reports, message, cc):
         """Notify reviewers on their unverified reports."""
         User = get_user_model()
         reviewers = User.objects.all_reviewers().filter(email__isnull=False)
         subject = '[Timed] Verification of reports'
         from_email = settings.DEFAULT_FROM_EMAIL
-        mails = []
+        connection = get_connection()
+        messages = []
+        match = re.match('[^@]+@[^@]+\.[^@]+', ', '.join(cc))
```
winged: Is this really necessary? If we're "validating" mail addresses, we should also tell the user if it failed..
I suggest leaving this out completely.

timed/reports/management/commands/notify_reviewers_unverified.py
```diff
@@ -80,13 +96,17 @@ def _get_unverified_reports(self, start, end):
...
+        if not match:
+            cc = ['']
```
winged: Why is this a list of an empty string? Would a empty list `[]` not be cleaner?

timed/reports/tests/test_notify_reviewers_unverified.py
```diff
@@ -27,9 +35,16 @@ def test_notify_reviewers(db, mailoutbox):
     ReportFactory.create(date=date(2017, 7, 1), task=task_no_work,
                          verified_by=reviewer_no_work)
 
-    call_command('notify_reviewers_unverified')
+    call_command(
+        'notify_reviewers_unverified',
+        '--cc={0}'.format(cc),
```
winged: We still need to test what happens if the user leaves out the parameters. not just the parameters being empty

JohnL17: I have added a test case for when the user leaves out the parameters.

PR #315, opened 2018-09-04
Change activity to accept changes when active

Yelinz:


timed/tracking/serializers.py
```diff
@@ -41,7 +41,8 @@ def validate(self, data):
 
         # validate that there is only one active activity
         activity = models.Activity.objects.filter(user=user)
-        if activity.filter(to_time__isnull=True) and to_time is None:
+        if (activity.filter(to_time__isnull=True) and
```
sliverc: what happens when I patch activity and set `to_time` to `None`? I guess we could have more than one activity with `to_time` `None` right?

Yelinz: I added an test and changed the condition

Yelinz: Changed it to how we discussed it

timed/tracking/serializers.py
```diff
@@ -39,12 +39,20 @@ def validate(self, data):
         to_time = data.get('to_time', instance and instance.to_time)
         user = instance and instance.user or data['user']
 
+        def validate_unique():
```
sliverc: Sounds quite generic and could be anything. Maybe `validate_running_activity` might be a better name? Or any other ideas?

Yelinz: Couldn't think of a better name, so i just used your suggestion

PR #330, opened 2018-10-24
Update the development setup to only use docker-compose

anehx:


sliverc:
Travis needs to be fixed so it doesn't use make targets.

PR #358, opened 2018-12-17
Add task row and task totals to workreport

Yelinz:


timed/reports/views.py
```diff
@@ -253,6 +260,17 @@ def _create_workreport(self, from_date, to_date, today, project, reports,
         table[13 + len(reports), 2].formula = 'of:=SUM(B13:B{0})'.format(
             str(13 + len(reports) - 1))
 
+        pos = 18 + len(reports)
+        for task in tasks.items():
```
sliverc: better:
`for task_name, task_total_hours in tasks.items()`

easier to read

timed/reports/views.py
```diff
@@ -253,6 +260,17 @@ def _create_workreport(self, from_date, to_date, today, project, reports,
...
+            table.insert_rows(pos, 1)
+            table[pos, 0] = Cell(
+                '{0}:'.format(task[0]),
```
sliverc: with above change this could simply be task_name

timed/reports/views.py
```diff
@@ -253,6 +260,17 @@ def _create_workreport(self, from_date, to_date, today, project, reports,
...
+                style_name=table[13 + len(reports), 0].style_name
+            )
+            table[pos, 2] = Cell(
+                task[1], style_name=table[13 + len(reports), 2].style_name
```
sliverc: task_total_hours here.

timed/reports/views.py
```diff
@@ -209,6 +209,7 @@ def _create_workreport(self, from_date, to_date, today, project, reports,
         tmpl = settings.WORK_REPORT_PATH
         doc = opendoc(tmpl)
         table = doc.sheets[0]
+        tasks = {}
```
sliverc: you could use defaultdict to avoid below if/else.

https://docs.python.org/3/library/collections.html#collections.defaultdict

timed/reports/views.py
Yelinz: Done

timed/reports/views.py
Yelinz: Done

timed/reports/views.py
Yelinz: Done

timed/reports/views.py
Yelinz: Done

sliverc:
@Yelinz :+1:

PR #396, opened 2019-02-19
Fix duplicated absences for supervisees

sbor23:
This commit fixes a bug where supervisees had their absences
duplicated by the number of supervisors they have.


sbor23:
Hope it makes more sense now

PR #397, opened 2019-02-20
Remove rest_condition for REST framework operators

sbor23:
Replace existing permission_classes composed using rest_condition
operator `C()` with native operators.
Introduce new Permission `IsNotDelete` as complement to `IsDeleteOnly`
because there's no negation allowed, only bitwise `&` and `|`.

Get rid of rest_condition dependency.

### DIFF
--- a/timed/permissions.py
+++ b/timed/permissions.py
@@ -28,6 +28,16 @@ class IsDeleteOnly(BasePermission):
         return self.has_permission(request, view)
 
 
+class IsNotDelete(BasePermission):
+    """Disallow delete method."""
+
+    def has_permission(self, request, view):
+        return request.method != "DELETE"
+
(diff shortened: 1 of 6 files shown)


timed/permissions.py
```diff
@@ -28,6 +28,16 @@ def has_object_permission(self, request, view, obj):
         return self.has_permission(request, view)
 
 
+class IsNotDelete(BasePermission):
+    """Allows only delete method."""
```
sliverc: comment needs to be adjusted.... although in this case the class name is quiet clear so might just also remove it.

sbor23:
> I assume you have done some manual testing as well that it still behaves the same? Just to be safe...

I made sure it's cover by existing tests, eg. `test_absence_delete_superuser()`.

PR #398, opened 2019-02-20
Disallow deletion of BillingType and CostCenter

sbor23:
Disallow deletion of BillingType and CostCenter if they are referenced
in a project and/or task.
This is propagated into the admin site as well.


sbor23:
I'm not sure if it's better usability to remove the "Delete" button in the admin site completely.
The alternative would be to leave it, this way the user would see this:
image

This way the user would have a hint on what tasks are referencing the to-be-deleted item.

sbor23:
I'm not sure if it's better usability to remove the "Delete" button in the admin site completely.
The alternative would be to leave it, this way the user would have a hint on what projects are referencing the to-be-deleted item.:
image

sbor23:
Changes to the admin site are purged. We stick to the default django warning which doesn't allow deletion.

PR #517, opened 2020-03-11
feat: show spent_billable time on projects

sbor23:


sbor23:
I think is intended and how "spent-time" works too. The PL wants to see how much is spent already NOW, without having to verify the reports before.

PR #527, opened 2020-03-24
fix: replace deprecated pagination middleware

sbor23:
/cc @aziiee


sbor23:
@fkm @aziiee This requires frontend changes, the meta field is now called `limit`:
https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html?highlight=paging#pagination

PR #537, opened 2020-04-07
fix(db): generate missing migration

anehx:


sbor23:
Whaat, was that missing since the update?
How did the CI not catch this...

PR #540, opened 2020-04-08
Revert "Match postgres version with production"

sbor23:
This reverts commit 923f8e23f2e1e30fd43d8ee09c6fbb0d18028013.
We don't want to use an EOL postgres version.
Also downgrading postgres is potentially dangerous, for example my
testsetup wasn't able to start because the db version wasn't supported
by the postgres 9.4 driver.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   db:
-    image: postgres:9.4
+    image: postgres:9.6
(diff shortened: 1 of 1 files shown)


sbor23:
We'll look into this once the new test and production environments on AKS are available.

sbor23:
This is not needed because we move to postgresql 12 in moment anyway.

PR #541, opened 2020-04-09
Performance fix for reports export

sbor23:
Make sure that reports-export does not crash.
Add the settings REPORTS_EXPORT_MAX_COUNT to optionally restrict
queryset to a certain length.
Move cost_center and billing_type logic to query annotations to make it
more performant.
Return to-be-exported list a generator expression to not fill up memory.

This also includes a fix where the cost center name used the
task.project.name instead of task.cost_center.name.

### DIFF
--- a/timed/tracking/views.py
+++ b/timed/tracking/views.py
@@ -1,7 +1,8 @@
 """Viewsets for the tracking app."""
 
 import django_excel
-from django.db.models import Q
+from django.conf import settings
+from django.db.models import Case, CharField, F, Q, Value, When
 from django.http import HttpResponseBadRequest
 from django.utils.translation import ugettext_lazy as _
 from rest_framework import exceptions, status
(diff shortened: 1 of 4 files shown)


sbor23:
The tests have shown that 400000+ reports work fine for csv but not for odt and xlsx.
Maybe there's more performance improvement in this front, since the bottle neck is now in the pyexcel-webio library (which isn't updated since 2+ years).

anehx:
Did you check the `/work-reports` endpoint as well?

sbor23:
On it

sbor23:
> Did you check the `/work-reports` endpoint as well?

See #543

timed/tracking/views.py
```diff
@@ -236,6 +211,41 @@ def export(self, request):
...
+        if (
+            settings.REPORTS_EXPORT_MAX_COUNT > 0
+            and queryset.count() > settings.REPORTS_EXPORT_MAX_COUNT
+        ):
+            return Response(
+                "Your request exceeds the maximum allowed entries ({0})".format(
+                    settings.REPORTS_EXPORT_MAX_COUNT
+                ),
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
open-dynaMIX, line 247: The message should be added to `django.po`. And maybe not only include how many entries are allowed, but also how many entries were included in the query.

sbor23: Haha didn't even realize there are "translations". Nobody touched it in more than 2 years so there's some more messages to check-in... :sweat_smile:

open-dynaMIX: :sweat_smile: 

Could you please add the missing translations strings in a separate commit.

sbor23: done

PR #542, opened 2020-04-09
chore: move to psycopg2 wheel

sbor23:
The psycopg2-bin package is intended for getting a dev setup running
quickly, but not for production use.
See: https://www.psycopg.org/docs/install.html#binary-install-from-pypi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ RUN wget https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for
 RUN apt-get update && apt-get install -y --no-install-recommends \
   libldap2-dev \
   libsasl2-dev \
+  libpq-dev \
   && rm -rf /var/lib/apt/lists/* \
   && mkdir -p /app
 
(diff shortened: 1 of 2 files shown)


open-dynaMIX:
Haha, sorry. I don't know how, but I was reading this PR the other way around. Please disregard my previous comment.

sbor23:
> We could consider using the wheel in dev and the source package in prod.

Regardless of the wrong-way-around: I was wondering if it's practicable to manage two separate Dockerfiles... But it would complicated the whole CI part, so let's not think about this.

open-dynaMIX:
This could be solved without an additional `Dockerfile`, just with separate dev and prod `requirements.txt`.

sbor23:
Ah yes but then we couldn't just treat the dev-packages as a superset of the prod packages.

open-dynaMIX:
Exactly. We would end up with three `requirements` files:

 - `requirements.txt` --> base
 - `requirements-dev` --> contains psycopg binary; inherits from base
 - `requirements-prod` --> contains psycopg source; inherits from base

In other projects, we're also not doing this. But IMO it's something we could adopt.

PR #543, opened 2020-04-09
feat: add setting to restrict work reports export size

sbor23:
The `WORK_REPORTS_EXPORT_MAX_COUNT` setting restricts work reports
export to the given count. This should fix cases where the (unfiltered)
queryset is too big to handle.

### DIFF
--- a/timed/reports/views.py
+++ b/timed/reports/views.py
@@ -7,8 +7,10 @@ from zipfile import ZipFile
 from django.conf import settings
 from django.db.models import F, Sum
 from django.db.models.functions import ExtractMonth, ExtractYear
-from django.http import HttpResponse, HttpResponseBadRequest
+from django.http import HttpResponse
 from ezodf import Cell, opendoc
+from rest_framework import status
+from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
 
 from timed.mixins import AggregateQuerysetMixin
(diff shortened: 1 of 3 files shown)


timed/reports/views.py
```diff
@@ -268,11 +268,16 @@ def _create_workreport(self, from_date, to_date, today, project, reports, user):
...
+            return HttpResponseBadRequest()
+
         # needed as we add items in reverse order to work report
         queryset = queryset.reverse()
-        count = queryset.count()
-        if count == 0:
-            return HttpResponseBadRequest()
+
+        if settings.WORK_REPORTS_EXPORT_MAX_COUNT > 0:
+            queryset = queryset[: settings.WORK_REPORTS_EXPORT_MAX_COUNT]
```
winged, line 279: From a business perspective, this could be problematic. You're silently removing a part from the output without the user noticing, which could lead to problems down the road, depending on what's being done with the report.

Instead, I'd prefer to have a 400 bad request with the message to filter the data some more to get below the max row count.

(The message of course would need to be shown in the frontend as well)

timed/reports/views.py
```diff
@@ -268,11 +268,16 @@ def _create_workreport(self, from_date, to_date, today, project, reports, user):
 
     def list(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
+
+        if queryset.count() == 0:
+            return HttpResponseBadRequest()
```
winged, line 273: I'm not sure that this "bad request" is warranted. In my opinion, an empty list is a valid output and shouldn't cause an error

timed/reports/views.py
sbor23: Valid point, that came to my mind too

timed/reports/views.py
sbor23: Since this endpoint is used for exports only it might make sense to prompt the user that nothing has been selected.
Doesn't make much sense to export an empty excel sheet.

timed/reports/views.py
sbor23: done

timed/reports/views.py
sbor23: Anyway I made this a restful response which is actually helpful for the frontend.

PR #547, opened 2020-04-15
fix(settings): make verified-email setting global

sbor23:


hairmare:
As discussed, it makes sense to have a global default for settings that we only rarely need to change. Having a default just for dev that doesn't apply to prod might make sense in some cases (ie. DJANGO_DEBUG). This is not one of those cases.

PR #560, opened 2020-05-26
Implement OIDC authentication

sbor23:
Drop LDAP and JWT based authentication in favour of OIDC.

The OIDC auth uses https://github.com/mozilla/mozilla-django-oidc with a customized authentication backend. The authentication backend checks the `userinfo` endpoint and `introspection` as a fallback (for confidental clients using client credentials grant in keycloak).


winged:
I think it's a good idea to sync with @open-dynaMIX regarding this - we've discussed a while back that we should try to unify all the OIDC backend code in our API projects - Timed, Emeis, Caluma etc. This might be a good chance to push this unification a bit forward.

I like that you're using a library for the OIDC stuff, but I'm not sure how much we actually gain from this. So probably it's best to discuss this in realtime :)

sbor23:
As discussed, I will write tests for this and see to actually get it running in a dev environment.
So we will stick to the package

timed/authentication.py
```diff
@@ -0,0 +1,109 @@
...
+    def get_or_create_user(self, access_token, id_token, payload):
+        """Verify claims and return user, otherwise raise an Exception."""
+
+        claims = self.get_userinfo_or_introspection(access_token)
+
+        # claims_verified = self.verify_claims(userinfo) or self.verify_claims(introspection)
+
+        # if not claims_verified:
+        #     msg = "Claims verification failed"
+        #     raise SuspiciousOperation(msg)
```
open-dynaMIX, line 64: Should be removed or used to verify the claims.

timed/authentication.py
```diff
@@ -0,0 +1,109 @@
...
+        last_name = claims.get(settings.OIDC_LASTNAME_CLAIM, "")
+        return self.UserModel.objects.create(
+            username=username, email=email, first_name=first_name, last_name=last_name
+        )
+
+    def get_username(self, claims):
+        try:
+            return claims[settings.OIDC_USERNAME_CLAIM]
+        except KeyError:
+            raise SuspiciousOperation("Couldn't find username claim")
```
open-dynaMIX, line 109: Did you intentionally not add an `update()` method?

timed/authentication.py
sbor23: did that

timed/authentication.py
sbor23: Yes, this is not a requirement for now. I left `create_user()` in there just in case (it's behind the settings flag), but the normal workflow is for admins to create the users manually

sbor23: FYI I removed the call to `self.update_user()` because it's a noop anyways.

PR #562, opened 2020-05-28
feat: extend docker-compose to include proxy, keycloak

sbor23:
Add keycloak and proxy to dev-setup.
Since we are moving to OIDC auth, we need an actual OIDC provider to
test in the local dev environment.
This commit includes all the necessary bits in the docker-compose setup
to get up and running locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,15 +18,21 @@ Timed timetracking software REST API built with Django
 After installing and configuring those requirements, you should be able to run the following
 commands to complete the installation:
 
+Add the `timed.local` entries to your hosts file:
+```bash
+echo "127.0.0.1 timed.local" | sudo tee -a /etc/hosts
+```
+
+Then just start the docker-compose setup:
 ```bash
 make start
 ```
 
-You can now access the API at http://localhost:8000/api/v1 and the admin interface at http://localhost:8000/admin/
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -30,5 +30,37 @@ services:
       - ENV=docker
       - STATIC_ROOT=/var/www/static
 
+  keycloak_db:
```
anehx, line 33: Why aren't you just using the existing db?

sbor23: Simply to have things separated. But since we have everything in the same docker-compose I might as well use the same db

README.md
```diff
@@ -36,6 +36,21 @@ To get the application working locally for development, make sure to create a fi
...
 
+To have the whole setup running, including the [timed-frontend](https://github.com/adfinis-sygroup/timed-frontend), we need some more strings attached.
+Add the `timed.local` entries to your host file:
+```bash
+echo "127.0.0.1 timed.local" | sudo tee -a /etc/hosts
+```
+
+Start the required services for a fully-fledged local development setup:
+```bash
+docker-compose --file docker-compose.optional.yml up -d
```
anehx, line 47: I don't see that file anywhere...

sbor23: Decided to include everything in the `docker-compose.yml` since a OIDC provider is needed anyways.

PR #586, opened 2020-07-13
Docker: Support basic uwsgi config via env variables

michaelimfeld:


cmd.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+sed -i 's/max-requests = .*/max-requests = '"${UWSGI_MAX_REQUESTS}"'/g' "${UWSGI_INI}"
+sed -i 's/harakiri = .*/harakiri = '"${UWSGI_HARAKIRI}"'/g' "${UWSGI_INI}"
+sed -i 's/processes = .*/processes = '"${UWSGI_PROCESSES}"'/g' "${UWSGI_INI}"
```
hairmare, line 5: You can also do multiple substitutions in one call to sed.
```suggestion
sed -i 's/max-requests = .*/max-requests = '"${UWSGI_MAX_REQUESTS}"'/g' \
    -i 's/harakiri = .*/harakiri = '"${UWSGI_HARAKIRI}"'/g' \
    -i 's/processes = .*/processes = '"${UWSGI_PROCESSES}"'/g' "${UWSGI_INI}"
```
You might also consider using something like envsubst.

PR #587, opened 2020-07-15
chore(docs): document oidc env vars, drop unused settings

sbor23:


timed/settings.py
```diff
@@ -207,34 +207,34 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
...
+    "DJANGO_OIDC_USERINFO_ENDPOINT",
+    default=default(f"{OIDC_DEFAULT_BASE_URL}/userinfo"),
 )
-OIDC_OP_TOKEN_ENDPOINT = env.str(
-    "OIDC_TOKEN_ENDPOINT", default=default(f"{OIDC_DEFAULT_BASE_URL}/token")
+OIDC_VERIFY_SSL = env.bool("DJANGO_OIDC_VERIFY_SSL", default=default(False, True))
+OIDC_CREATE_USER = env.bool("DJANGO_OIDC_CREATE_USER", default=False)
+
+OIDC_USERNAME_CLAIM = env.str(
+    "DJANGO_OIDC_USERNAME_CLAIM", default="preferred_username"
```
open-dynaMIX, line 221: Usually we use `sub` as default, but it can be configured in the client. If I understand it right, in Keycloak, `sub` is the default for access tokens, while `preferred_username` is the default for id tokens.

timed/settings.py
```diff
@@ -207,34 +207,34 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 
 OIDC_DEFAULT_BASE_URL = "http://timed.local/auth/realms/timed/protocol/openid-connect"
```
open-dynaMIX, line 208: Why is this not configurable?

sbor23: Good point!
It was initially only used for the dev setup, but definitely makes sense to expose this to have sane defaults for the userinfo and introspection endpoint.

timed/settings.py
sbor23: changed

timed/settings.py
sbor23: fixed

timed/settings.py
open-dynaMIX: Sorry, I saw it after approving it. This has not been addressed.

sbor23: Oops forgot to push it.

open-dynaMIX: This needs also to be changed in the tests.

PR #613, opened 2020-08-14
fix: increase uwsgi buffer-size

sbor23:
solves the problem of query-string limitations


sbor23:
Additional options are:
* max-requests-delta should help to even out worker restarts so they don't restart at the same time.
* smaller `max-requests` could prevent memory-leaks from growing too big.
* `max-worker-lifetime` could be an option too especially for fixing memory-leaks on not-so-frequently-used instances (eg. staging). But it seems a little dangerous as it can kill workers mid-request...

PR #634, opened 2020-10-02
feat: introduce code owners file

winged:
Let's see how this works out. This here is more of an experiment than
a production setting. If it gains traction, we'll add it to more projects
in the future

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,3 @@
+# Code owners for the Timed backend. We include our backend dev team here.
+# Since this is a split project (backend/frontend) it's rather simple
+* @adfinis-sygroup/dev-backend


CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Code owners for the Timed backend. We include our backend dev team here.
+# Since this is a split project (backend/frontend) it's rather simple
+* @adfinis-sygroup/dev/backend
```
sbor23, line 3: Does the slashed notion have any functional meaning to it? Never seen that before

winged: That's there to reference (nested) teams

sbor23: Ok I can't see such a nested example in [the docs](https://docs.github.com/en/enterprise/2.15/user/articles/about-code-owners) but fine with me.

sbor23 (approved):

CODEOWNERS
winged: Yeah you're right - nested teams are still referenced via `org/team-name` without nesting. Changed the file accordingly

winged: (... and renamed the teams as well for consistency)

hairmare:
I set up repo permissions for the @adfinis-sygroup/dev-backend group but the repo didn't seem to pick up on them yet.

PR #651, opened 2020-10-19
bump postgresql, typo

### LINKED ISSUE #538: Upgrade PostgreSQL to a non EOL version
As per 923f8e23f2e1e30fd43d8ee09c6fbb0d18028013 we seem to be using a PostgreSQL version that was EOLed February 13, 2020.

We should upgrade to a version that is not EOL, basically all versions >9.4 are still in support.

See here for the available versions and their support policy: https://www.postgresql.org/support/versioning/

Once updated the  Helm chart for timed also needs updating. In fact this issue surfaced during updating/implementing the chart: https://github.com/adfinis-sygroup/timed-chart/pull/9#issuecomment-610854977

sbor23:
resolves #538

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   db:
-    image: postgres:9.4
+    image: postgres:12.4
     ports:
       - 5432:5432
     volumes:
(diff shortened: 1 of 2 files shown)


winged (approved):
:+1:

PR #663, opened 2020-11-23
📝 readme: add steps to have a working development environment with users

marcaurele:
Add documentation on the required steps to have a working development environment with the keycloak integration in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,12 @@ Then you'll be able to login in the Django admin interface [http://timed.local/a
 
 ### Adding a user
 
-If you want to add other users with different roles, add them in the Keycloak interface (as they would be coming from your LDAP directory). You will also have to correct their location in the Django admin interface as it is not correctly set for the moment. Head to [http://timed.local/admin/](http://timed.local/admin/) to after having perform 1 login with the uuser. You should see that new user in the `Employment -> Users`. Click on the user and scroll down to the `Employments` section to set a `Location`. Save the user and you should now see the _Timed_ interface correctly under that account.
+If you want to add other users with different roles, add them in the Keycloak interface (as they would be coming from your LDAP directory).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,6 +49,28 @@ If you have existing users from the previous LDAP authentication, you want to ad
...
+Password (again): 
+Password changed successfully for user 'admin'
+```
+
+Then you'll be able to login in the Django admin interface [http://timed.local/admin/](http://timed.local/admin/).
+
+
+### Adding a user
+
+If you want to add other users with different roles, add them in the Keycloak interface (as they would be coming from your LDAP directory). You will also have to correct their location in the Django admin interface as it is not correctly set for the moment. Head to [http://timed.local/admin/](http://timed.local/admin/) to after having perform 1 login with the uuser. You should see that new user in the `Employment -> Users`. Click on the user and scroll down to the `Employments` section to set a `Location`. Save the user and you should now see the _Timed_ interface correctly under that account.
```
sbor23, line 72:
```suggestion
If you want to add other users with different roles, add them in the Keycloak interface (as they would be coming from your LDAP directory).
You will also have to correct their employment in the Django admin interface as it is not correctly set for the moment.
Head to [http://timed.local/admin/](http://timed.local/admin/) after having perform a first login with the user.
You should see that new user in the `Employment -> Users`.
Click on the user and scroll down to the `Employments` section to set a `Location`.
Save the user and you should now see the _Timed_ interface correctly under that account.
```

sbor23 (changes requested):
I'm a bit unsure if this should go under `Development` section, but more information is generally better. Thanks for this!

marcaurele:
> I'm a bit unsure if this should go under `Development` section, but more information is generally better. Thanks for this!

I found `Development` better than `Installation`.  I was also thinking about a section `Initial development setup` maybe.

sbor23 (approved):

PR #664, opened 2020-11-24
fix: add test data users to keycloak config

Yelinz:


sbor23:
Would you mind adding a django fixture with the mentioned users as well?
Most importantly with an employment, otherwise the frontend will break.

Yelinz:
A fixture already exists with exactly those users https://github.com/adfinis-sygroup/timed-backend/blob/master/timed/fixtures/test_data.json

sbor23 (approved):

PR #670, opened 2020-12-08
feat: add billed to project and tracking

Yelinz:
needs #668 

Implements flags on project and report to mark them as billed.
Changes the permissions on report to allow editing verified reports but not billed and verified


timed/tracking/tests/test_report.py
```diff
@@ -93,8 +94,10 @@ def test_report_intersection_full(auth_client):
 
 def test_report_intersection_partial(auth_client):
     user = auth_client.user
-    ReportFactory.create(review=True, not_billable=True, comment="test")
+    report = ReportFactory.create(review=True, not_billable=True, comment="test")
```
sbor23, line 97: Would you mind using the `report_factory` fixture for this and do it consistently in for the whole file?
I started doing this already, so we have (eventual) consistency throughout the projects.

sbor23: Also, set `billed=True` right away.

timed/tracking/tests/test_report.py
```diff
@@ -282,6 +286,18 @@ def test_report_list_filter_editable_supervisor(auth_client):
     assert len(json["data"]) == 3
 
 
+def test_report_list_filter_billed(auth_client):
+    report = ReportFactory.create(billed=True)
+
+    url = reverse("report-list")
+
+    response = auth_client.get(url)
```
sbor23, line 294: Please test some filters too, as the test name suggests

timed/tracking/tests/test_report.py
```diff
@@ -333,6 +349,41 @@ def test_report_create(auth_client):
     assert json["data"]["relationships"]["task"]["data"]["id"] == str(task.id)
 
 
+def test_report_create_billed(auth_client):
+    """Should create a new report and automatically set the user."""
+    user = auth_client.user
+    project = ProjectFactory.create(billed=True)
+    task = TaskFactory.create(project=project)
```
sbor23, line 356: same with `task_factory` pls

timed/tracking/tests/test_report.py
```diff
@@ -668,6 +719,28 @@ def test_report_reset_verified_by_reviewer(auth_client):
         }
     }
 
+    url = reverse("report-detail", args=[report.id])
+    response = auth_client.patch(url, data)
+    assert response.status_code == status.HTTP_200_OK
```
sbor23, line 724: maybe check that the data actually changed.

timed/tracking/tests/test_report.py
```diff
@@ -1019,3 +1092,51 @@ def test_report_update_bulk_review_and_verified(
...
+            "type": "reports",
+            "id": report.id,
+            "attributes": {"billed": True},
+        }
+    }
+
+    url = reverse("report-detail", args=[report.id])
+
+    response = auth_client.patch(url, data)
+    print(vars(response))
```
sbor23, line 1124: debug statement?

timed/tracking/tests/test_report.py
```diff
@@ -1019,3 +1092,51 @@ def test_report_update_bulk_review_and_verified(
...
+        "data": {
+            "type": "reports",
+            "id": report.id,
+            "attributes": {"billed": True},
+        }
+    }
+
+    url = reverse("report-detail", args=[report.id])
+    response = auth_client.patch(url, data)
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
sbor23, line 1142: A test case of successful (bulk-) update would be of interest too.

timed/permissions.py
```diff
@@ -134,3 +134,11 @@ class IsNotTransferred(BasePermission):
 
     def has_object_permission(self, request, view, obj):
         return not obj.transferred
+
+
+class IsBilledAndVerfied(BasePermission):
```
sbor23, line 139: This permission is only ever used in negated form, so building the negated version would make more sense.
There is also a typo: `IsNotBilledAndVerified`

sbor23 (changes requested):

sbor23:
Btw the first commit still shows up in the PR, I don't know if gh is at fault here of if you have diverging commits called "feat(projects): add currency fields to task and project" ?

They have different commit hashes so I propose you rebase.

timed/tracking/tests/test_report.py
Yelinz: The factory doesnt seem to be setting billed if I set it in the parameters, do you have any clue as why that could be?

sbor23: Have you tried to `rm -rf .pytest_cache`?

Yelinz:
Fixed the various problems and rebased to master

timed/permissions.py
```diff
@@ -134,3 +134,10 @@ class IsNotTransferred(BasePermission):
 
     def has_object_permission(self, request, view, obj):
         return not obj.transferred
+
+
+class IsNotBilledAndVerfied(BasePermission):
+    """Allows access only to billed objects."""
```
sbor23, line 140: description needs an update too.

Yelinz: Done

timed/tracking/tests/test_report.py
```diff
@@ -39,8 +38,11 @@ def test_report_list(auth_client):
     assert json["meta"]["total-time"] == "01:00:00"
 
 
-def test_report_intersection_full(auth_client):
-    report = ReportFactory.create()
+def test_report_intersection_full(
+    auth_client,
+    report_factory,
```
sbor23, line 43: just `report` would be enough, no need to create it manually using the factory.

timed/tracking/models.py
```diff
@@ -105,11 +106,16 @@ def save(self, *args, **kwargs):
...
         However, the duration must at least be 15 minutes long.
+
+        Sets the billed state to the billed state of the project.
         """
         self.duration = timedelta(
             seconds=max(15 * 60, round(self.duration.seconds / (15 * 60)) * (15 * 60))
         )
 
+        if not self.pk or not self.billed:
+            self.billed = self.task.project.billed
```
sbor23, line 117: This marks all newly created reports as `billed` if the project is `billed`.
What about existing reports?
* Should they get marked as `billed` if the project is updated (set to `billed=True`)?
In that case we would need to overwrite the `Project.save()` method too.

* Should they get marked as `billed` if they are updated after `project` is updated?

Yelinz: 1. It wasn't specified to do that automatically, so no.
2. Yes, if they get moved or updated on a billed project it should also mark them

timed/tracking/tests/test_report.py
sbor23: Looks good now.
Just FYI, you could use `report_factory(foo=bar)` instead of `report_factory.create(..)`.

timed/tracking/tests/test_report.py
sbor23: This is not resolved yet, see leftovers below.

timed/permissions.py
sbor23: Thanks!

timed/tracking/models.py
sbor23: Alright, for 2 I'm pretty sure this doesn't work yet. Please add a test case where the project is set as billed and an update on a report (text field or so) yields a `billed=True`.

```suggestion
        if not self.billed:
            self.billed = self.task.project.billed
```

Yelinz: Added the test and changed it so it updates the billed flag

timed/permissions.py
```diff
@@ -134,3 +134,10 @@ class IsNotTransferred(BasePermission):
 
     def has_object_permission(self, request, view, obj):
         return not obj.transferred
+
+
+class IsNotBilledAndVerfied(BasePermission):
+    """Allows access only to not billed and not verfied objects."""
+
+    def has_object_permission(self, request, view, obj):
+        return not (obj.billed is True and obj.verified_by_id is not None)
```
sbor23, line 143:
```suggestion
        return not obj.billed or obj.verified_by_id is None
```

sbor23 (approved):

sbor23:
Good job!

PR #681, opened 2020-12-30
chore(ci): remove codecov

sbor23:


winged (approved):

sbor23:
The `.coverage` file should be availble in the action host through the docker volume map, but it seems codecov cannot pick it up. Maybe because pytest is executed as root and thus the file is owned by root?

winged:
> The `.coverage` file should be availble in the action host through the docker volume map, but it seems codecov cannot pick it up. Maybe because pytest is executed as root and thus the file is owned by root?

If you run `pytest` within the container, are you sure the `.coverage` is visible "outside"? The "override" compose file suggests the workdir is mounted to `/app`, but I think this may be the problem.

sbor23:
> If you run `pytest` within the container, are you sure the `.coverage` is visible "outside"? The "override" compose file suggests the workdir is mounted to `/app`, but I think this may be the problem.

It is visible on my local dev setup. Maybe an additional step to copy the file to the host is needed...

.github/workflows/test.yml
```diff
@@ -13,14 +13,18 @@ jobs:
...
+          docker-compose exec -T backend pip install -r requirements-dev.txt
       - name: Lint the code
         run: |
           docker-compose exec -T backend black --check .
           docker-compose exec -T backend flake8
           docker-compose exec -T backend ./manage.py makemigrations --check --dry-run --no-input
       - name: Run pytest
         run: docker-compose exec -T backend pytest --no-cov-on-fail --cov --create-db -vv
+      - name: Copy .coverage
+        run: docker cp $(docker ps -q backend):/app/.coverage .coverage
```
winged, line 26: This should be `docker-compose ps -q backend` I think.

Ugly workaround, but I don't see a better way either

sbor23: Thanks!

sbor23:
Turns out that codecov doesn't support pythons `.coverage`, I decided to remove codecov instead because it's not really useful if we enforce 100%

winged:
> Turns out that codecov doesn't support pythons `.coverage`, I decided to remove codecov instead because it's not really useful if we enforce 100%

lol, so that was the issue all along... :facepalm:  :grin:

PR #693, opened 2021-02-26
chore(deps): upgrade django to 3.1.7

trowik:
upgrade python to 3.9 and bump other dependencies


trowik:
the tests are failing because of a deprecation in the urls

sbor23:
Can you please run `.manage.py makemigrations` and see if that helps? Seems like a couple of PRs have an issue with a missing migration somehow

sbor23:
Ah well, the issue comes from another package, see https://github.com/mozilla/mozilla-django-oidc/pull/371 .
We need to temporarily disable raising warnings as errors in pytest. Please comment the line `    error::DeprecationWarning` and add a comment about the issue.

setup.cfg
```diff
@@ -48,7 +48,8 @@ addopts = --reuse-db --randomly-seed=1521188767 --randomly-dont-reorganize
 env=
     DJANGO_OIDC_USERNAME_CLAIM=sub
 filterwarnings =
-    error::DeprecationWarning
+    # comment is needed beceause of https://github.com/mozilla/mozilla-django-oidc/pull/371
+    # error::DeprecationWarning
     error::PendingDeprecationWarning
```
sbor23, line 53: welp seems like you have to do this one too...

open-dynaMIX: Sorry for chiming in. Please only ignore specific DeprecationWarnings. This can be achieved with a pattern, like [here](https://github.com/adfinis-sygroup/mySAGW/blob/main/api/setup.cfg#L44).

sbor23: much better suggestion!

setup.cfg
```diff
@@ -50,6 +50,10 @@ env=
 filterwarnings =
     error::DeprecationWarning
     error::PendingDeprecationWarning
+    ignore:Using a non-boolean value for an isnull lookup is deprecated, use True or False instead.:django.utils.deprecation.RemovedInDjango40Warning
```
sbor23, line 53: where does this come from?

sbor23 (approved):
LGTM!

PR #709, opened 2021-04-08
fix: add billable column and calculate not billable time

trowik:
The new formula still needs recalculation, which may be annoying, but can be enabled in the Tools to recalculate on File Load or manually by pressing F9.
I couldn't find another solution for this problem.


timed/reports/views.py
```diff
@@ -265,6 +268,13 @@ def _create_workreport(self, from_date, to_date, today, project, reports, user):
             str(13 + len(reports) - 1)
         )
 
+        # calculate location of total not billable hours as insert rows moved it
+        table[
+            13 + len(reports) + len(tasks) + 1, 2
+        ].formula = 'of:=SUMIF(C13:C{0};"nein";B13:B{0})'.format(
+            str(13 + len(reports) - 1)
+        )
+
```
sbor23, line 277: refactor this into (local) function as it does exactly the same as the lines above, except for a `+1` and `C` instead of `B`.

sbor23 (changes requested):

timed/reports/views.py
trowik: It adds a new formula, but I can still refactor this into a function.

sbor23: Ah true, nevermind then.

sbor23 (approved):
LGTM

trowik:
test coverage 99.98% :roll_eyes:

winged:
> The new formula still needs recalculation, which may be annoying, but can be enabled in the Tools to recalculate on File Load or manually by pressing F9.
> I couldn't find another solution for this problem.

Not entirely sure if possible, but maybe you could do the calculation in python as well, and fill in both formula and value into the cell?

trowik:
Could find a way to fill in both the formula and the value.

timed/reports/views.py
```diff
@@ -226,10 +226,13 @@ def _create_workreport(self, from_date, to_date, today, project, reports, user):
 
             hours = report.duration.total_seconds() / 60 / 60
             table["B13"] = Cell(hours, style_name=float_style)
-
-            table["C13"] = Cell(report.user.get_full_name(), style_name=text_style)
-            table["D13"] = Cell(report.task.name, style_name=text_style)
-            table["E13"] = Cell(report.comment, style_name=text_style)
+            if report.not_billable:
+                table["C13"] = Cell("nein", style_name=text_style)
```
winged, line 230: Since we're doing all reports in english now anyway, I'd suggest using either "yes"/"no" or something neutral (like placing an "X" mark)

timed/reports/tests/test_work_report.py
```diff
@@ -21,7 +21,12 @@ def test_work_report_single_project(auth_client, django_assert_num_queries):
     project = ProjectFactory.create(customer=customer, name="Project/")
     task = TaskFactory.create(project=project)
     ReportFactory.create_batch(
-        10, user=user, verified_by=user, task=task, date=date(2017, 8, 17)
+        10,
```
winged, line 24: I suggest creating some billable ones, and some non-billables, so we have both to test.

winged (changes requested):
Looking good! Just some minor remarks

timed/reports/views.py
sbor23: Agree, but make sure it's consistent with the naming of the column. In which case a switch of logic would be necessary.

`not report.not_billable` -> "X" mark

trowik: I have to translate the work report to english anyways so I'll do it in another MR.

winged (approved):

PR #711, opened 2021-04-09
fix: add custom forms for supervisor and supervisee inlines

trowik:
labels of foreign keys for the supervisor and supervisee inlines
needed to be changed

### DIFF
--- a/timed/employment/admin.py
+++ b/timed/employment/admin.py
@@ -16,7 +16,38 @@ from timed.forms import DurationInHoursField
 admin.site.disable_action("delete_selected")
 
 
+class SupervisorForm(forms.ModelForm):
+    """Custom form for the supervisor admin."""
+
+    # Change the label of the supervisor through table attribute to_user
+    to_user = forms.ModelChoiceField(
+        queryset=models.User.objects.all(), label=_("supervised by")
+    )
+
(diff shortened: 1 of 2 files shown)


sbor23 (approved):
Looks good, but please shorten the commit message so it fits into one line.

winged:
LGTM as well, but agree with @sbor23 regarding the commit message.

Use the first line as a sort of "subject" (what it does), followed by an empty line, and then some text explaining the change (how is it done, why, etc).

See https://chris.beams.io/posts/git-commit/ and https://www.conventionalcommits.org/en/v1.0.0/

winged:
LGTM as well, but agree with @sbor23 regarding the commit message.

Use the first line as a sort of "subject" (what it does), followed by an empty line, and then some text explaining the change (how is it done, why, etc).

See https://chris.beams.io/posts/git-commit/ and https://www.conventionalcommits.org/en/v1.0.0/

trowik:
Sure, is something like now better?

winged:
> Sure, is something like now better?

Better, and I'll merge it in a second, but I'd have swapped the sentences: If you look at `git log --oneline`, I want to see that the label changed (which was the driving idea) not how it was done (adding custom forms)

PR #719, opened 2021-04-13
feat: export metrics with django-prometheus

sbor23:
~~Note that currently the url `/prometheus/metrics` requires authentication. This can be setup using a confidential client given in the documentation.~~

~~Auth-free metrics is out of scope for the moment.~~

Edit: disregard above note, metrics can be scrapped from `/prometheus/metrics` without authentication.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help start test shell loaddata
+.PHONY: help start stop test shell flush loaddata
 .DEFAULT_GOAL := help
 
 help:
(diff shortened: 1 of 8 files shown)


hairmare (approved):

Makefile
```diff
@@ -1,18 +1,23 @@
-.PHONY: help start test shell loaddata
+.PHONY: help start stop test shell flush loaddata
```
winged, line 1: A cleaner pattern would be to add the `.PHONY: foo` atop of every make target instead. Helps with not forgetting it.

timed/settings.py
```diff
@@ -27,7 +27,7 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 DATABASES = {
     "default": {
         "ENGINE": env.str(
-            "DJANGO_DATABASE_ENGINE", default="django.db.backends.postgresql"
+            "DJANGO_DATABASE_ENGINE", default="django_prometheus.db.backends.postgresql"
```
winged, line 30: Intrusive plugin is intrusive

winged (approved):
Minor comment regarding cleanliness, but feel free to merge anyway, it's not that important

PR #722, opened 2021-04-22
fix(workreport): template layout with correct line-wrapping

sbor23:
Fix:
* "description" is line-wrapping, not "task"
* "billable" is centered
* compress short columns so there's more space for description

Before, the "description column would wrap to another page when exporting as pdf / print.


winged (approved):
LGTM

PR #726, opened 2021-04-28
feat: add user assignement and external employees

trowik:
add user assignement for customers, projects and tasks with the role resource, manager and reviewer
removed reviewers from django admin
add migration command for reviewers
add external employments
filter customers and projects for external employees


trowik:
**Don't merge yet!**

winged (approved):
LGTM!

timed/projects/views.py
```diff
@@ -17,15 +27,29 @@ class CustomerViewSet(ReadOnlyModelViewSet):
...
+
         :return: The customers
         :rtype:  QuerySet
         """
-        return models.Customer.objects.prefetch_related("projects")
+        user = self.request.user
+        current_employment = Employment.objects.get_at(user=user, date=date.today())
+
+        if not current_employment.is_external:  # pragma: no cover
+            return models.Customer.objects.prefetch_related("projects")
```
winged, line 39: The prefetch should also happen for external users. This should be done outside the if/else

timed/projects/views.py
```diff
@@ -54,8 +84,16 @@ class ProjectViewSet(PreloadIncludesMixin, ReadOnlyModelViewSet):
...
-        return queryset.select_related("customer", "billing_type", "cost_center")
+        """Get only assigned projects, if an employee is external."""
+        user = self.request.user
+        current_employment = Employment.objects.get_at(user=user, date=date.today())
+
+        if not current_employment.is_external:  # pragma: no cover
+            queryset = super().get_queryset()
+            return queryset.select_related("customer", "billing_type", "cost_center")
+        else:
+            queryset = models.Project.objects.filter(assignees=user)
```
winged, line 95: This should also use `super().get_queryset()`. Minor refactoring required

timed/projects/views.py
```diff
@@ -83,3 +121,14 @@ def filter_queryset(self, queryset):
             self.ordering = None
 
         return super().filter_queryset(queryset)
+
+    def get_queryset(self):
+        """Get only assigned tasks, if an employee is external."""
+        user = self.request.user
+        # current_employment = Employment.objects.get_at(user=user, date=date.today())
```
winged, line 128: Remove commented-out code

timed/tracking/serializers.py
```diff
@@ -153,6 +153,41 @@ def validate(self, data):
...
+                | Q(
+                    project__project_assignees__user=user,
+                    project__project_assignees__is_manager=True,
+                )
+                | Q(
+                    project__customer__customer_assignees__user=user,
+                    project__customer__customer_assignees__is_manager=True,
+                )
+            ).exists()
+            and self.context["request"].method == "POST"
```
winged, line 185: I'm not sure the `POST` check here is required.

Also, if you reorder the conditions so that the `is_external` check comes first, Python can shortcut the expression and avoid the query if user is internal.

winged (changes requested):
Nice work, but still some things to clean up

timed/tracking/serializers.py
trowik: `POST` is here, because only external `is_resource` should be able to create new reports, reviewers should only update them.
The error message is wrong there.

timed/projects/views.py
trowik: My bad :sweat_smile:

timed/projects/views.py
trowik: Ah yes, makes sense

timed/projects/serializers.py
```diff
@@ -100,6 +116,34 @@ def get_root_meta(self, resource, many):
 
         return {}
 
+    def validate(self, data):
+        if self.context["request"].method == "POST":
```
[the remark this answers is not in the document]
trowik: Of course, `POST` isn't needed here

timed/projects/serializers.py
```diff
@@ -100,6 +116,34 @@ def get_root_meta(self, resource, many):
...
+                    project_assignees__user=user,
+                    project_assignees__is_reviewer=True,
+                )
+                | Q(
+                    customer__customer_assignees__user=user,
+                    customer__customer_assignees__is_manager=True,
+                )
+                | Q(
+                    customer__customer_assignees__user=user,
+                    customer__customer_assignees__is_reviewer=True,
```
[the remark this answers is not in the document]
trowik: Yes, you can create new tasks in frontend if you go to the projects tab

timed/projects/serializers.py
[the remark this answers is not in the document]
trowik: It is needed to differentiate between needed role for specific method.
I added a description that explains it. Hope it helps.

timed/projects/serializers.py
```diff
@@ -100,6 +116,54 @@ def get_root_meta(self, resource, many):
...
+        Check if the external employee is a manager or reviewer on the corresponding
+        project or customer when he wants to create a new task.
+
+        Check if the external employee is a manager or reviewer on the task or
+        the corresponding project or customer when he wants to update the task.
+        """
+        request = self.context["request"]
+        user = request.user
+        if (
+            request.method in ["PATCH", "PUT"]
```
trowik, line 131: Couldn't find a way without checking the request method. Open to any suggestions.

timed/permissions.py
```diff
@@ -99,21 +112,35 @@ def has_permission(self, request, view):
         if not super().has_permission(request, view):  # pragma: no cover
             return False
 
-        if request.method not in SAFE_METHODS:
-            return request.user.reviews.exists()
-
-        return True
+        if request.method in ["PATCH", "GET", "POST"]:
+            return True
```
winged, line 116: No need to check for the access method. This is done by combining the permission with for example `IsNotDelete` etc.

timed/permissions.py
```diff
@@ -141,3 +168,150 @@ class IsNotBilledAndVerfied(BasePermission):
...
+
+
+class IsInternal(IsAuthenticated):
+    """Allows access only to internal employees."""
+
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):  # pragma: no cover
+            return False
+
+        if request.method not in SAFE_METHODS:
```
winged, line 180: Drop the request method check. This permission should only check if the user is internal, not anything else. You can combine the permission with other permissions to achieve the actual check you need

timed/permissions.py
```diff
@@ -141,3 +168,150 @@ class IsNotBilledAndVerfied(BasePermission):
...
+
+
+class IsExternal(IsAuthenticated):
+    """Allows access only to external employees."""
+
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):  # pragma: no cover
+            return False
+
+        if request.method not in SAFE_METHODS:
```
winged, line 210: Same here - check for "external" state only

timed/permissions.py
```diff
@@ -141,3 +168,150 @@ class IsNotBilledAndVerfied(BasePermission):
...
+
+
+class IsManager(IsAuthenticated):
+    """Allows access only to assignees with manager role."""
+
+    def has_permission(self, request, view):
+        if not super().has_permission(request, view):  # pragma: no cover
+            return False
+
+        if request.method not in SAFE_METHODS:
```
winged, line 241: No method check here either.

Also, you should `return` boolean in every case, which doesn't happen in this method.

timed/permissions.py
```diff
@@ -141,3 +168,150 @@ class IsNotBilledAndVerfied(BasePermission):
...
+            return (
+                request.user.customer_assignees.filter(is_manager=True).exists()
+                or request.user.project_assignees.filter(is_manager=True).exists()
+                or request.user.task_assignees.filter(is_manager=True).exists()
+                or False
```
winged, line 246: `or False` doesn't do anything. Just leave it out

timed/permissions.py
```diff
@@ -141,3 +168,150 @@ class IsNotBilledAndVerfied(BasePermission):
...
+                    )
+                    | Q(
+                        project__customer__customer_assignees__user=user,
+                        project__customer__customer_assignees__is_manager=True,
+                    )
+                )
+                .exists()
+            )
+        else:  # pragma: no cover
+            return False
```
winged, line 272: Throw an exception, similar to what you're doing in `IsReviewer`

winged (changes requested):
We're getting there!

A few things are still left to do, but we are getting there.

timed/permissions.py
trowik: done

winged (approved):

winged:
Nice work!

PR #733, opened 2021-05-04
fix(redmine): template formatting

sbor23:


trowik (approved):
Looks good to me, if the auth header isn't necessary

sbor23:
> Looks good to me, if the auth header isn't necessary

It's not, the `key` does the trick already. Not sure if this is a leftover from an old version of the redmine library or it was redundant from the beginning.

PR #734, opened 2021-05-04
fix(runtime): use gunicorn instead of uwsgi

winged:
Also, use 8 workers by default instead of 4.

This is partly experimental, and partly due to us having tons of performance problems on prod, which could very well be due to the poor UWSGI config.


sbor23 (approved):
Thanks for the PR!

Skimming the documentation there might be a couple of flags we might want to check at some point, but we can do that using the `GUNICORN_CMD_ARGS` at a later point.

* https://docs.gunicorn.org/en/latest/settings.html#threads
* https://docs.gunicorn.org/en/latest/settings.html#timeout

PR #762, opened 2021-06-24
fix(redmine): update template formatting

trowik:


winged (commented):
Looks good, however I'd like to see the output a bit more compact.

We could potentially shorten the "column" a bit by concatenating the output. Right now, it's (roughly) as follows:

```
2021-06-25 2.25h [NB]  [Rev] comment ....
2021-06-25 2.25h [NB]        comment ....
2021-06-25 2.25h             comment ....
2021-06-25 2.25h       [Rev] comment ....
```

Whereas, if we compact it a bit, or ignore the tabular output altogether, we'd gain a few more precious characters:

```
2021-06-25 2.25h [NB,R] comment ....
2021-06-25 2.25h [NB] comment ....
2021-06-25 2.25h comment ....
2021-06-25 2.25h [R] comment ....
```

tk0e:
I fully agree with @winged and would prefer his proposal to the original design.

trowik:
done in #773

PR #772, opened 2021-08-06
feat: use assignees instead of reviewers

trowik:
use assignees with reviewer role instead of project reviewers
remove reviewers table


timed/projects/filters.py
```diff
@@ -44,7 +43,7 @@ def filter_has_manager(self, queryset, name, value):
             )
         )
 
-    def filter_has_reviewer(self, queryset, name, value):
+    def filter_has_reviewer(self, queryset, name, value):  # pragma: no cover
```
winged, line 46: Please write tests instead of using `# pragma: no cover`. Especially for parts that contain non-trivial code. Same for the filter method above.

winged (changes requested):
Please write tests for your filters - they're not trivial pieces of code. Otherwise LGTM

timed/projects/filters.py
```diff
@@ -24,7 +24,40 @@ class ProjectFilterSet(FilterSet):
...
+                ).values("project_id"),
+            )
+            | Q(
+                customer_id__in=models.CustomerAssignee.objects.filter(
+                    is_manager=True, user_id=value
+                ).values("customer_id"),
+            )
+        )
+
+    def filter_has_reviewer(self, queryset, name, value):
```
JohnL17, line 46: Isn't the method `filter_has_reviewer` exactly the same as the method `filter_has_manager`?

trowik: No, `filter_has_manager` filters assignees with the manager role and `filter_has_reviewer` filters assignees with the reviewer role. One has `is_manager=True` and the other `is_reviewer=True`.

JohnL17: Ah ok I can see it now :wink:

JohnL17:
Overall looks good to me.

JohnL17 (approved):

Issue #776, opened 2021-08-13
Static files don't load on test aks


trowik:
Static files for django admin don't load on the test aks.
Maybe it's an issue with the `STATIC_ROOT` not being correct.
[Gunicorn PR](https://github.com/adfinis-sygroup/timed-backend/commit/a9c9788c327d27b7f18b348db341217716ab6fa3#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

PR #777, opened 2021-08-17
feat(deployment): serve static files

winged:
This is against django best practices, but this is mainly a DRF application,
and the only static files we serve are for the Django-Admin, with a limited
user base. Thus, we can serve static files without too much negative impact

### DIFF
--- a/timed/urls.py
+++ b/timed/urls.py
@@ -1,5 +1,7 @@
 """Root URL mapping."""
 
+from django.conf import settings
+from django.conf.urls.static import static
 from django.contrib import admin
 from django.urls import include, re_path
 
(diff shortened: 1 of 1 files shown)


winged:
Fixes #776

open-dynaMIX (approved):
:neutral_face: 

Would it make sense to use [whitenoise](http://whitenoise.evans.io/en/stable/django.html) for this?

winged:
> neutral_face
> 
> Would it make sense to use [whitenoise](http://whitenoise.evans.io/en/stable/django.html) for this?

Thanks! IMHO, Whitenoise is not worth the effort though - given the use case. If we did have more users and more static content, I'd agree

PR #779, opened 2021-08-23
fix(projects): add manager role to project assignees

aziiee:
Until the introduction of assignees, a reviewer could verify
reports and manage tasks of a project. Now, this has changed and
tasks can only be managed by assignees who have the manager role.


winged (changes requested):
This is not gonna work, as the migration 0013 removes all assignees.

This needs to happen during migration `0012_migrate_reviewers_to_assignees`. I suggest adding the operation to that file instead.

Of course, we will need to re-run this on Test, but that's another issue

winged (approved):

PR #782, opened 2021-09-01
feat(employment): add is_accountant flag for user

trowik:


winged (approved):
Looks good! I assume the check for this flag in the report viewset coming in a separate PR?

trowik:
If by "check in report viewset" you meant the permissions for accountants to update reports, then yes

winged (approved):
Looks good!

PR #793, opened 2021-09-28
fix(redmine): fix total hours calculation

trowik:
total_hours was calculated for all reports during updated-range and not for total duration of the project
reverted the change in the command added in  #733

### DIFF
--- a/timed/redmine/tests/test_redmine_report.py
+++ b/timed/redmine/tests/test_redmine_report.py
@@ -83,3 +83,38 @@ def test_redmine_report_invalid_issue(db, freezer, mocker, capsys, report_factor
 
     _, err = capsys.readouterr()
     assert "issue 1000 assigned" in err
+
+
+def test_redmine_report_calculate_total_hours(
+    db, freezer, mocker, task, report_factory
+):
+    redmine_instance = mocker.MagicMock()
+    issue = mocker.MagicMock()
+    redmine_instance.issue.get.return_value = issue
+    redmine_class = mocker.patch("redminelib.Redmine")
+    redmine_class.return_value = redmine_instance
+
+    freezer.move_to("2017-07-15")
+    reports = report_factory.create_batch(10, task=task)
+
+    freezer.move_to("2017-07-24")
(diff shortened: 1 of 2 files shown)


winged (changes requested):
Looks good, and I see how it fixes the issue. However, for a change that fixes such a nasty issue, I suggest adding a test for it.

timed/redmine/tests/test_redmine_report.py
```diff
@@ -83,3 +83,40 @@ def test_redmine_report_invalid_issue(db, freezer, mocker, capsys, report_factor
...
+
+    freezer.move_to("2017-07-15")
+    reports = report_factory.create_batch(10, task=task)
+
+    freezer.move_to("2017-07-24")
+    reports_last_seven_days = report_factory.create_batch(10, task=task)
+
+    total_hours_last_seven_days = 0
+    for report in reports_last_seven_days:
+        total_hours_last_seven_days = (
```
winged, line 105: Use `+=` operator to make it more readable

winged (changes requested):
Looks good, thanks for adding the test.

To make the code more readable, consider using `queryset.aggregate(Sum('duration'))` or at least the `+=` operator as suggested

timed/redmine/tests/test_redmine_report.py
trowik: Ah yes, forgot about that

timed/redmine/tests/test_redmine_report.py
```diff
@@ -83,3 +83,40 @@ def test_redmine_report_invalid_issue(db, freezer, mocker, capsys, report_factor
...
+            total_hours_last_seven_days + report.duration.total_seconds() / 3600
+        )
+
+    total_hours = 0
+    for report in reports + reports_last_seven_days:
+        total_hours = total_hours + report.duration.total_seconds() / 3600
```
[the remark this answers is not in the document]
trowik: done

winged (approved):

PR #796, opened 2021-10-05
fix(permissions): rename IsNotBilledAndVerified permissions

trowik:
rename permission to depict its functionality


aziiee (approved):
LGTM

PR #799, opened 2021-10-13
fix(tracking): set correct value for billed flag on reports

trowik:
Use project's billed flag on new reports (e.g. when a project was paid in advance).
If a report is moved to a different project, task or customer, set the billed flag on the report according to the project the report was moved to.
Removed IsNotBilledOrVerified permission.
Tests for the billed flag are already implemented.

Following needs clarification with the admin:
- [x]  Setting the billed flag on reports to false on already billed projects. Should it be allowed or not?
- [x] What should happen with the reports, if a project is set from billed to not billed? As of now, the reports will only be updated, if anything is changed on them (by manually saving a report).

### DIFF
--- a/timed/projects/models.py
+++ b/timed/projects/models.py
@@ -2,7 +2,12 @@
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
 from djmoney.models.fields import MoneyField
+
+from timed.tracking.models import Report
 
 
 class Customer(models.Model):
@@ -242,3 +247,21 @@ class TaskAssignee(models.Model):
     is_resource = models.BooleanField(default=False)
     is_reviewer = models.BooleanField(default=False)
     is_manager = models.BooleanField(default=False)
+
+
+@receiver(pre_save, sender=Project)
+def update_billed_flag_on_reports(sender, instance, **kwargs):
+    """Update billed flag on all reports from the updated project.
+
+    Only update reports if the billed flag on the project was changed.
+    Setting the billed flag to True on a project in Django Admin should set
+    all existing reports to billed=True. Same goes for setting the flag to billed=False.
+    The billed flag should primarily be set in frontend.
+    This is only a quicker way for the accountants to update all reports at once.
+    """
+    # check whether the project was created or is being updated
+    if instance.pk:
+        if instance.billed != Project.objects.get(id=instance.id).billed:
+            Report.objects.filter(Q(task__project=instance)).update(
+                billed=instance.billed
+            )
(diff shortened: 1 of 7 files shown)


trowik:
Removing the billed flag is necessary for moving reports from an already billed project (e.g. SLA, which are paid in advance and are already billed/flag on project set to billed). So it makes sense to move reports to an already billed project.
The process of billing is new. Individual steps, as of how it should work in the end, aren't documented or known (yet). I discussed with the admin, how the process should look like (more or less):
- Primarily the billed flag should be set in frontend by accountants after the reports are verified, unless the project is paid in advance, in which case the reports should already be set to billed.
- Setting the billed flag to True on a project should set all reports (next and previous) to billed=True. Same goes for setting the flag to billed=False.

timed/projects/models.py
```diff
@@ -242,3 +247,18 @@ class TaskAssignee(models.Model):
...
+
+    Only update reports if the billed flag on the project was changed.
+    """
+    # check whether the project was create or is being updated
+    if instance.pk:
+        if instance.billed != Project.objects.get(id=instance.id).billed:
+            for report in Report.objects.filter(Q(task__project=instance)):
+                if report.billed != instance.billed:
+                    report.billed = instance.billed
+                    report.save()
```
winged, line 264:
```suggestion
            Report.objects.filter(Q(task__project=instance)).update(billed=instance.billed)
```

No need to iterate over every report - it's much easier to just `.update()` the queryset, and faster too. Let the DB do the work

timed/projects/models.py
```diff
@@ -242,3 +247,18 @@ class TaskAssignee(models.Model):
     is_resource = models.BooleanField(default=False)
     is_reviewer = models.BooleanField(default=False)
     is_manager = models.BooleanField(default=False)
+
+
+@receiver(pre_save, sender=Project)
+def update_billed_flag_on_reports(sender, instance, **kwargs):
+    """Update billed flag on all reports from the updated project.
+
+    Only update reports if the billed flag on the project was changed.
```
winged, line 256: I would like to see a docstring that explains the logic/intention behind the code, so that readers in the future understand why we're doing this. Just copy a fragment from the PR here

winged (changes requested):
Looks good, some optimisation / docstrings to do however.

Still not entirely convinced about the business case, but I'm not gonna block because of that.

timed/projects/models.py
trowik: Didn't know about that, thx :)

timed/projects/models.py
trowik: Added a short explanation. I hope it helps understanding the logic behind it. Let me know if it's enough or not.

winged (approved):

PR #800, opened 2021-11-02
feat(reports): change column for total hours for tasks

trowik:
Moved the column of total hours spent on task and total hours spent in general to the right.
workreport_changed_column


winged (approved):
LGTM!

PR #802, opened 2021-11-10
fix(fixtures): add user with accountant role

aziiee:
This will also add a command in Makefile which allows to start a
dbshell.


trowik (approved):
LGTM

PR #803, opened 2021-11-10
feat: include user in ReportIntersection

aziiee:
- Check if certain Report objects are from the same user
- Fix filtering of editable and non-editable reports for accountants


winged (approved):
Looks good

winged:
Some tests are still missing the new attributes

PR #804, opened 2021-11-11
fix(filters): billing should not affect finish state of a report

aziiee:
This is especially the case with projects that are billed in advance.


aziiee:
cc @trowik

winged (approved):

PR #805, opened 2021-11-11
feat(settings): add DATA_UPLOAD_MAX_NUMBER_FIELDS to env

trowik:
This value should be customizable as there are big projects that have a lot of tasks, tasks and project assignees. In these cases the amount of fields exceeds the default value of 1000, which makes saving any changes on that project impossible.
I'll add the variable to our helm chart as well, so we can change it at any time.


winged (approved):
LGTM, however setting this to an unreasonable large amount can make us vulnerable to DOS (in theory). But given that we allow authenticated requests only anyway,  [we could also just set it to `None` and have Django skip the field count check](https://docs.djangoproject.com/en/3.2/ref/settings/#data-upload-max-number-fields)

PR #807, opened 2021-11-23
fix(tracking): update reviewer filter

trowik:
only show reports in which user is responsible for reviewing
e.g. if user is project assignee with reviewer role, don't show
reports in which according task assignee with reviewer role is set


winged (approved):
Nice work! This definitely makes more sense than before

PR #808, opened 2021-11-24
feat: update visibility to not depend on employment

trowik:
update visibility in Report-, User-, Customer-, Project- and TaskViewSet
update employment check in ReportSerializer

### DIFF
--- a/timed/employment/views.py
+++ b/timed/employment/views.py
@@ -24,7 +24,7 @@ from timed.permissions import (
     IsSupervisor,
     IsUpdateOnly,
 )
-from timed.projects.models import Task
+from timed.projects.models import CustomerAssignee, Task
 from timed.tracking.models import Absence, Report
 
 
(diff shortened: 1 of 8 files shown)


timed/employment/views.py
```diff
@@ -53,30 +53,31 @@ class UserViewSet(ModelViewSet):
...
             "employments", "supervisees", "supervisors"
         )
 
-        if current_employment.is_external:
-            assigned_tasks = Task.objects.filter(
-                Q(task_assignees__user=user, task_assignees__is_reviewer=True)
-                | Q(
-                    project__project_assignees__user=user,
-                    project__project_assignees__is_reviewer=True,
+        if user.employments.all().exists():
```
winged, line 60: This checks whether a user *had* an employment *at any time*, which is incorrect. You need to call the `get_at()` method as you do within the `if` statement anyway to figure out if the user has an employment. You can use `try/except` to handle the case where the user doesn't have an active employment.

timed/employment/views.py
```diff
@@ -53,30 +53,31 @@ class UserViewSet(ModelViewSet):
...
+            current_employment = models.Employment.objects.get_at(
+                user=user, date=datetime.date.today()
+            )
+            if current_employment.is_external:
```
winged, line 64:
```suggestion
            if not current_employment or current_employment.is_external:
```

winged (changes requested):
Nice work so far, but the situation where user doesn't have an employment (Current or not) needs to be improved

timed/employment/views.py
```diff
@@ -53,32 +53,36 @@ class UserViewSet(ModelViewSet):
...
-            )
-
-            return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
 
-        return queryset
+                return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
+            return queryset
+        except models.Employment.DoesNotExist:
+            if CustomerAssignee.objects.filter(user=user).exists():
+                return queryset.filter(Q(id=user.id))
```
trowik, line 84: I'll add a is_customer check later with the implementation of the new role

timed/employment/views.py
trowik: Makes sense. I refactored the code to use `try/except`

timed/employment/views.py
trowik: That doesn't really work, because I get an `DoesNotExist` error if there's no employment for the specific user at the current date.

winged (approved):
Looks good now!

PR #811, opened 2021-12-09
Send mails on order creation

trowik:
moved acknowledged validation to OrderView

### DIFF
--- a/timed/subscription/notify_admin.py
+++ b/timed/subscription/notify_admin.py
@@ -1,0 +1,53 @@
+import datetime
+
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives, get_connection
+from django.template.loader import get_template, render_to_string
+
+
+def prepare_and_send_email(project, order_duration):
+    template_txt = get_template("notify_accountants_order.txt")
+    from_email = settings.DEFAULT_FROM_EMAIL
+    connection = get_connection()
+    messages = []
+
+    customer = project.customer
+
+    duration = datetime.datetime.strptime(order_duration, "%H:%M:%S")
+    hours_added = datetime.timedelta(
+        hours=duration.hour, minutes=duration.minute, seconds=duration.second
+    )
+    hours_total = hours_added + project.estimated_time
+
+    subject = f"Customer Center Credits/Reports: {customer.name} has ordered {hours_added} hours."
+
+    body_txt = template_txt.render(
+        {
+            "customer": customer,
+            "project": project,
+            "hours_added": hours_added,
+            "hours_total": hours_total,
+        }
+    )
+    body_html = render_to_string(
+        "notify_accountants_order.html",
+        {
+            "customer": customer,
+            "project": project,
+            "hours_added": hours_added,
+            "hours_total": hours_total,
+        },
+    )
+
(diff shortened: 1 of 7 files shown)


trowik:
@winged could you review the last commit? Will rebase after #810 is merged

timed/subscription/templates/notify_accountants_order.html
```diff
@@ -0,0 +1,17 @@
+<b>Aufladung von {{hours_added}} Stunde/n</b>
```
winged, line 1: I suggest translating the template to english.

Also, nobody is probably ever going to buy a single hour, so no need to use the singular here

timed/subscription/views.py
```diff
@@ -50,6 +56,64 @@ class OrderViewSet(viewsets.ModelViewSet):
...
+        """Override so we can issue emails on creation."""
+        # check if order is acknowledged and created by admin/accountant
+        if (
+            request.method == "POST"
+            and request.data.get("acknowledged")
+            and not (request.user.is_accountant or request.user.is_superuser)
+        ):
+            raise ValidationError("User can not create confirmed orders!")
+
+        template_txt = get_template("notify_accountants_order.txt")
```
winged, line 69: Refactor: I suggest moving the whole emailing code into it's own module or at least a separate function/method.

winged (changes requested):
Some minor changes, otherwise looks good

timed/subscription/templates/notify_accountants_order.html
trowik: done

timed/subscription/views.py
trowik: done

trowik:
changed the name of env variable for customer-center email

timed/subscription/notify_admin.py
```diff
@@ -0,0 +1,56 @@
...
+    project = Project.objects.get(id=request.data.get("project")["id"])
+    customer = project.customer
+
+    duration = datetime.datetime.strptime(request.data.get("duration"), "%H:%M:%S")
+    hours_added = datetime.timedelta(
+        hours=duration.hour, minutes=duration.minute, seconds=duration.second
+    )
+    hours_total = hours_added + project.estimated_time
+
+    subject = f"Customer Center Credits/Reports: {customer.name} hat {hours_added} Stunde/n bestellt."
```
winged, line 25: German text

timed/subscription/notify_admin.py
```diff
@@ -0,0 +1,56 @@
+import datetime
+
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives, get_connection
+from django.template.loader import get_template, render_to_string
+
+from timed.projects.models import Project
+
+
+def prepare_and_send_email(request):
```
winged, line 10: This is business logic, not request handling. Avoid passing in the request object. Instead, pass in the project model directly. This way, the function can be run from other contexts as well if needed

winged (changes requested):

timed/subscription/notify_admin.py
trowik: Had to pass order duration as well

winged (approved):

PR #812, opened 2021-12-09
fix(subscription): fix visibility on subscription-projects for customers

trowik:
Only show own (customer-visible) projects for customers

### DIFF
--- a/timed/subscription/tests/test_subscription_project.py
+++ b/timed/subscription/tests/test_subscription_project.py
@@ -1,10 +1,13 @@
 from datetime import timedelta
 
+import pytest
 from django.urls import reverse
-from rest_framework.status import HTTP_200_OK
+from rest_framework.status import HTTP_200_OK, HTTP_404_NOT_FOUND
 
+from timed.employment.factories import EmploymentFactory
 from timed.projects.factories import (
     BillingTypeFactory,
+    CustomerAssigneeFactory,
     CustomerFactory,
     ProjectFactory,
     TaskFactory,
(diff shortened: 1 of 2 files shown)


timed/subscription/tests/test_subscription_project.py
```diff
@@ -64,3 +65,20 @@ def test_subscription_project_detail(auth_client):
...
+    user = auth_client.user
+    CustomerAssigneeFactory.create(user=user, customer=customer, is_customer=True)
+
+    url = reverse("subscription-project-list")
+
+    response = auth_client.get(url)
+    assert response.status_code == HTTP_200_OK
+
+    json = response.json()
+    assert len(json["data"]) == 1
```
winged, line 84: I suggest checking not only the length, but also the content of the returned data (ID of the returned project should be the exact one that we're allowed to see)

timed/subscription/views.py
```diff
@@ -27,7 +28,16 @@ class SubscriptionProjectViewSet(viewsets.ReadOnlyModelViewSet):
...
+        user = self.request.user
+        queryset = Project.objects.filter(archived=False, customer_visible=True)
+        if CustomerAssignee.objects.filter(user=user, is_customer=True).exists():
+            return queryset.filter(
+                Q(
+                    customer__customer_assignees__user=user,
+                    customer__customer_assignees__is_customer=True,
+                )
+            )
+        return queryset
```
winged, line 40: This returns all the projects if you're *NOT* a customer assignee. Not exactly sure this is what you want

winged: I suggest adding a check for user being employee before returning the full queryset

winged (approved):

PR #813, opened 2021-12-09
fix: update fixtures and keycloak config

aziiee:
This is especially need for the new customer-center changes


winged (approved):
Guess it's OK - can't really verify all the changes

PR #819, opened 2021-12-10
fix(docker-compose): add hint for keycloak config export

aziiee:


winged (approved):
Good call documenting this :-)

PR #821, opened 2021-12-14
feat(settings): add CORS_ALLOWED_ORIGINS to env

trowik:
Customer Center needs to make requests to timed so we have to allow CORS from some domains that we can configure.


aziiee:
Thanks!

winged (approved):
LGTM!

PR #823, opened 2021-12-16
fix(subscriptions/notify_admin): use dateutils parser to prevent an error

velrest:
Using strptime throws an error if the passed date is over 23 hours.
So instead we use dateutils.parser. We also pass the day value to the mail template.


winged (approved):
Good catch!

PR #829, opened 2021-12-17
fix(subscription/notify_admin): prevent invalid addition of datetime …

velrest:
…and int

### DIFF
--- a/timed/subscription/notify_admin.py
+++ b/timed/subscription/notify_admin.py
@@ -21,8 +21,10 @@ def prepare_and_send_email(project, order_duration):
         minutes=duration.minute,
         seconds=duration.second,
     )
-    estimated_time = 0 if project.estimated_time is None else project.estimated_time
-    hours_total = hours_added + estimated_time
+
+    hours_total = hours_added
+    if project.estimated_time is not None:
+        hours_total += project.estimated_time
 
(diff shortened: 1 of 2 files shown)


timed/subscription/notify_admin.py
```diff
@@ -21,8 +21,10 @@ def prepare_and_send_email(project, order_duration):
         minutes=duration.minute,
         seconds=duration.second,
     )
-    estimated_time = 0 if project.estimated_time is None else project.estimated_time
-    hours_total = hours_added + estimated_time
+
+    hours_total = hours_added
+    if project.estimated_time is not None:
+        hours_total += project.estimated_time
```
winged, line 27: Why not be fully explicit and do `hours_total += datetime.timedelta(hours=project.estimated_time)`?

velrest: I just tested and this will also throw an error if `project.estimated_time` is `None`.

winged: Yeah the `None` check is required in every case

velrest: But isnt estimated_time anyway a timedelta if its set?

winged: Hmm, I thought it was an `int`, but you're right. Should be `timedelta` or `None`

winged (approved):

PR #843, opened 2022-01-11
fix(reports): add reviewer hierarchy in notify_reviewers_unverified

trowik:
I used the same approach as in the reviewer filter for reports.


winged (approved):
Nice work!

PR #847, opened 2022-01-27
fix(projects): change permissions and visibility for billing types

trowik:


timed/projects/views.py
```diff
@@ -55,11 +56,27 @@ class BillingTypeViewSet(ReadOnlyModelViewSet):
...
+
+        current_employment = user.get_active_employment()
+
+        if current_employment is None:
+            if models.CustomerAssignee.objects.filter(
+                user=user, is_customer=True
+            ).exists():
+                return queryset.filter(projects__customer_visible=True)
+        elif not current_employment.is_external:
+            return queryset
```
winged, line 79: I suggest changing the `if` structure to an early return for the "normal" case. If I see it correctly, the code doesn't properly handle the case where you have an "external" employment - in this case, it returns `None` (should be `queryset.none()`).

Then, the customer assignee should probably be limited to the billing types for the "own" projects only - not that the billing types are actually that critical.

So, we end up with a rough implementation like this:

```python
if current_employment:
    return queryset.none() if current_employment.is_external else queryset

return queryset.filter(
    projects__customer_visible=True,
    projects__customer__assignee__user=user,
    projects__customer__assignee__is_customer=True
)
```

timed/projects/tests/test_billing_type.py
```diff
@@ -3,21 +3,27 @@
...
 @pytest.mark.parametrize(
-    "is_employed, is_customer_assignee, is_customer, status_code",
+    "is_employed, is_customer_assignee, is_customer, project__customer_visible, expected, status_code",
     [
-        (False, True, False, HTTP_403_FORBIDDEN),
-        (False, True, True, HTTP_403_FORBIDDEN),
-        (True, False, False, HTTP_200_OK),
-        (True, True, False, HTTP_200_OK),
-        (True, True, True, HTTP_200_OK),
+        (False, True, False, False, 0, HTTP_403_FORBIDDEN),
```
winged, line 11: This doesn't test the case if the user is external. The current code would likely HTTP/500 in that situation

winged (changes requested):

timed/projects/views.py
trowik: Permissions don't allow access for users with external employment. That's why I didn't including handling of that case.
However there is a potential case in which external employment might be relevant: in a case where a customer has an (external) employment.
I'm not sure though whether that's a probable case.

winged (approved):

PR #849, opened 2022-02-08
fix(subscription): fix parser and notifications for orders

trowik:
notify only on orders created by customers
fix parser to correctly display purchased timed (for durations over 24h the parser didn't show correct numbers)


winged (approved):
LGTM :+1:

PR #851, opened 2022-02-22
fix(tracking): allow updating billed reports

trowik:
updating billed reports should be allowed, as there are projects, that are billed in advance

### DIFF
--- a/timed/tracking/serializers.py
+++ b/timed/tracking/serializers.py
@@ -117,6 +117,16 @@ class ReportSerializer(TotalTimeRootMetaMixin, ModelSerializer):
         """Only owner is allowed to change duration."""
         return self._validate_owner_only(value, "duration")
 
+    def validate_billed(self, value):
+        """Only accountants may bill reports."""
+        if self.instance is not None:
+            if not self.context["request"].user.is_accountant and (
+                self.instance.billed != value
+            ):
+                raise ValidationError(_("Only accountants may bill reports."))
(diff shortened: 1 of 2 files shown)


winged (approved):
LGTM

anehx (approved):
Nice work!

trowik:
Complexity is too high :unamused: 
Any ideas on how to lower it? Or should I just increase the max-complexity for flake8?

anehx:
@trowik use the `validate_billed` method for validating this. You can use `validate_{field_name}` for all validations that are specific to one field only.

trowik:
re-requested review for the validate_billed change

timed/tracking/serializers.py
```diff
@@ -117,6 +117,15 @@ def validate_duration(self, value):
         """Only owner is allowed to change duration."""
         return self._validate_owner_only(value, "duration")
 
+    def validate_billed(self, value):
+        """Only accountatns may bill reports."""
+        if self.instance is not None:
+            user = self.context["request"].user
+            if not user.is_accountant and (self.instance.billed != value):
```
anehx, line 124:
```suggestion
            if not self.context["request"].user.is_accountant and (self.instance.billed != value):
```

anehx (changes requested):
Otherwise LGTM :+1:

anehx (approved):

PR #871, opened 2022-07-19
feat(tracking): reject reports

trowik:
reviewers can reject reports, if they were reported on the wrong task, project or customer
send mails when a report was rejected
only reviewer may reject reports
automatically un-reject reports when they were moved by the owner


timed/tracking/models.py
```diff
@@ -184,3 +187,14 @@ class Meta:
...
 
         unique_together = ("date", "user")
+
+
+@receiver(pre_save, sender=Report)
+def update_rejected_on_reports(sender, instance, **kwargs):
+    """Unreject report when the task changes."""
+    # Check if the report is being created or updated
+    if instance.pk:
+        if instance.rejected:
```
winged, line 197:
```suggestion
    if instance.pk and instance.rejected:
```

timed/tracking/models.py
```diff
@@ -184,3 +187,14 @@ class Meta:
...
+            report = Report.objects.get(id=instance.id)
+            if report.task != instance.task:
```
winged, line 199: If you compare task_id instead of task, you get one or two DB queries less

timed/tracking/serializers.py
```diff
@@ -250,6 +260,7 @@ class ReportBulkSerializer(Serializer):
     not_billable = serializers.BooleanField(required=False, allow_null=True)
     billed = serializers.BooleanField(required=False, allow_null=True)
     verified = serializers.BooleanField(required=False, allow_null=True)
+    rejected = serializers.BooleanField(required=False, allow_null=True)
```
winged, line 263: Why allow_null? It's either True or False, no null reqired

winged (changes requested):
Some minor things to clean up, otherwise LGTM, from an implementation perspective.

However, I think the feature is rather incomplete in it's current requirements. What we really want is to tell people not only *that* a report was rejected, but also (optionally) why, and how to correct it. Maybe we should think ahead a bit and add a `review_comment` field or something similar, to be filled and displayed where needed?

trowik:
The comment on rejects isn't really necessary, as there are only two cases in which a report would be rejected. (Or at least I see only two cases):
1. The task, project or customer is wrong on the rejected report. In this case the reviewer won't exactly know which task/project/customer is correct, only that it shouldn't be reported on that specific task/project/customer. This would lead in most cases to a message similar to "Please change the task/project/customer. The report doesn't belong here."
2. The comment is inapproriate or not detailed (e.g. for customer visible reports). But this is rather an edge case and doesn't require the comment feature. (yet)
And if there really is a demand for a comment section after testing/release, we can still add it later on.

winged (approved):

Issue #872, opened 2022-07-19
rename default branch from master to main


hairmare:
i think we should do this all over the place, most modern projects have already switched

PR #873, opened 2022-07-19
chore(ci): fix building on legacy default branch

hairmare:
I also created #872 for tracking the fact that we should move to the new default anyway


trowik (approved):
:+1:

PR #874, opened 2022-07-20
feat: remaining effort

trowik:
#877 
Track remaining effort of selected tasks.
Manager can activate ```remaining effort tracking``` on a project, which enables setting of the estimated remaining effort.
Remaining effort is set on ```Report``` model and the field ```most_recent_remaining_effort``` on ```Task``` gets the most recent value. (Is updated via event handler)


winged:
As discussed: I suggest putting the `remaining_effort` field into the `Report` model instead. This will not only simplify the permissions (Regular employees will never need to edit `Task`s), but, more importantly, will give us a timeline of how the value changed over time (Burndown chart). Otherwise, you'll only ever have one value, and you don't know when it was updated and at which point in time it had any meaning.

When editing a report in the frontend, once a task is selected, the remaining effort field will need to be updated from the task, so the user knows what the last value was and can then update it according to their estimate.

Since we're going to be analyzing the remaining effort from the project perspective, we will potentially have a huge amount of calculation to do when someone requests a list of projects (Which happens a lot!). Therefore, it makes sense to denormalize this into the `Task` and `Project` models:

I suggest adding a field `most_recent_remaining_effort` on the `Task` and `Project` models, which would then be updated in a `post_save` hook on the `Report`. (The name is to clearly denote that this is always the most recent value, not the value at a given time).

timed/tracking/signals.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@receiver(post_save, sender=Report)
+def update_most_recent_remaining_effort(sender, instance, **kwargs):
+    """Update remaining effort on task, if remaining effort tracking is active."""
+    if instance.task.project.remaining_effort_tracking:
+        task = Task.objects.get(id=instance.task_id)
+        if instance == Report.objects.filter(task=task).last():
+            task.most_recent_remaining_effort = instance.remaining_effort
+            task.save()
```
winged, line 24:
```suggestion
        task = instance.task
        last_report = task.reports.order_by('date').last()
        if last_report == instance:
            task.most_recent_remaining_effort = last_report.remaining_effort
            task.save()
        project = task.project
        # Note: This following bit was just written from memory and needs to be verified..
        sum_remaining = project.tasks.all().annotate(sum_remaining=Sum('remaining_effort')).get().remaining_effort
        project.most_recent_remaining_effort = sum_remaining
        project.save()
```

winged (changes requested):
I think we should update the project as well... I've added a code suggestion to point you in the right direction

timed/reports/views.py
```diff
@@ -129,7 +132,10 @@ def get_queryset(self):
         queryset = Report.objects.all()
         queryset = queryset.values("task")
         queryset = queryset.annotate(duration=Sum("duration"))
-        queryset = queryset.annotate(pk=F("task"))
+        queryset = queryset.annotate(
+            pk=F("task"),
+            most_recent_remaining_effort=F("task__most_recent_remaining_effort"),
+        )
```
winged, line 138: Do we really need this here?

Or maybe we should change this queryset to start from `Task` instead. This would make tons of more sense and we'd avoid a lot of hacky stuff:

```python
queryset = Task.objects.all().annotate(duration=Sum('reports__duration'))
```

timed/tracking/serializers.py
```diff
@@ -136,6 +136,14 @@ def validate_rejected(self, value):
 
         return value
 
+    def validate_remaining_effort(self, value):
+        """Only update remaining effort when tracking is active on the corresponding project."""
+        if not self.instance.task.project.remaining_effort_tracking:
+            raise ValidationError(
+                "Remaining effort tracking is not active on this project!"
+            )
+        return value
```
winged, line 145: Instead of rejecting the API call, I'd just silently set it to `None` instead. No need to tell the user, as they're probably unable to do something about it

timed/tracking/signals.py
```diff
@@ -0,0 +1,34 @@
...
+        last_report = task.reports.order_by("date").last()
+        if instance == last_report:
+            task.most_recent_remaining_effort = instance.remaining_effort
+            task.save()
+        project = task.project
+        total_remaining_effort = (
+            project.tasks.all()
+            .aggregate(sum_remaining=Sum("most_recent_remaining_effort"))
+            .get("sum_remaining")
+        )
```
winged, line 32: Does this work? it's still a queryset, so `.get()` will return a record, not a value... right?

winged (changes requested):
Looks good! Some questions / minor issues to cleanup

timed/tracking/signals.py
trowik: Yes, ```total_remaining_effort``` is a value.
But I can write it as you proposed:
```project.tasks.all().annotate(sum_remaining=Sum('most_recent_remaining_effort')).get().most_recent_remaining_effort```
It didn't work for me one time so I refactored it to use aggregate, but I found the mistake and both ways return a value.

timed/tracking/serializers.py
trowik: Not so sure about that. At least they should get a ```Somehting gone wrong``` message when they can't save the remaining effort, otherwise the user might think it worked.

timed/reports/views.py
trowik: Definitely. This change will come with the next PR. I'm already working on it.
This is just a temporary solution so we can use it for frontend adjustments/development.

timed/tracking/serializers.py
winged: Guess it depends on frontend implementation. Feel free to keep this one then

winged (approved):

PR #879, opened 2022-08-04
chore: rename github orga from adfinis-sygroup to adfinis

hairmare:
I wanted to touch the workflow since i just recently built it but decided it's simpler to do all of it in one go.

We want this fixed so the next release is published via CI. While GitHub did add a redirect to the container image registry, support for following such redirect is extremely spotty and pretty much a docker only feature.


trowik (approved):
Thank you very much. Looks good. :+1:

PR #885, opened 2022-08-22
fix(reports): refactor statistics

trowik:
set queryset starting point from Report to Task, so we can get Tasks/Projects/Customers with no Reports
add new FilterSet for Customer-, Project- and TaskStatistics


trowik:
#886

PR #886, opened 2022-09-02
feat(filters): allow filtering of tasks and reports in statistics

winged:
When filtering statistics, we face the problem that we want to
have a list of all tasks, and count the reported hours. Due to the
way Django works, the Sum of reported hours would trigger an (unfiltered)
subquery or Join, so our filters don't apply.

Therefore, we now have a "split" pseudo-queryset, which contains separate
querysets for tasks and reports. It then applies the filters to EITHER
of the querysets as needed (depending on prefix), then combines them
at the end of the filtering phase, using a subquery.

### DIFF
--- a/timed/employment/admin.py
+++ b/timed/employment/admin.py
@@ -4,6 +4,7 @@ import datetime
 
 from django import forms
 from django.contrib import admin
+from django.contrib.admin.widgets import AutocompleteSelect
 from django.contrib.auth.admin import UserAdmin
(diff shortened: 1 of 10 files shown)


winged (commented):
Looking good, but you need to find another reviewer, as I cannot approve it, coming from my fork :-)

fugal-dy (approved):
Nice work!

trowik:
messed up commit history, so I opened a new PR #891

PR #887, opened 2022-09-08
feat: add searchable dropdowns for user lists in admin

trowik:
add new user filter (internal employment)


winged (changes requested):
Please do not mix multiple unrelated aspects in one commit (or even one PR). Code LGTM, but please split it into two MRs (one for the filter, one for the admin stuff)

trowik:
#890 is the filter MR

PR #891, opened 2022-10-05
Filter stats

trowik:
When filtering statistics, we face the problem that we want to
have a list of all tasks, and count the reported hours. Due to the
way Django works, the Sum of reported hours would trigger an (unfiltered)
subquery or Join, so our filters don't apply.

Therefore, we now have a "split" pseudo-queryset, which contains separate
querysets for tasks and reports. It then applies the filters to EITHER
of the querysets as needed (depending on prefix), then combines them
at the end of the filtering phase, using a subquery.

### DIFF
--- a/timed/reports/filters.py
+++ b/timed/reports/filters.py
@@ -5,7 +5,14 @@ from timed.projects.models import CustomerAssignee, ProjectAssignee, TaskAssigne
 from timed.projects.models import Task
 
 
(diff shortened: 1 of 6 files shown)


winged (approved):
LGTM, but due to my involvement in this PR, I suggest someone else should review this as well (@open-dynaMIX or @fugal-dy?)

fugal-dy (approved):

PR #892, opened 2022-10-05
refactor(app-server): replace uWSGI with Django Hurricane

fugal-dy:
Django Hurricane offers some advantage over uwsgi deployment in a cloud setup by providing dedicated probe endpoints for k8s to decide on pod health and app status


cmd.sh
```diff
@@ -4,6 +4,5 @@ GUNICORN_WORKERS="${GUNICORN_WORKERS:-8}"
```
winged, line 4: There's some leftover gunicorn env var above

winged (changes requested):
Looks good, but some minor cleanups / open questions remain

cmd.sh
fugal-dy: true

docker-compose.override.yml
```diff
@@ -14,7 +14,9 @@ services:
       - DJANGO_OIDC_USERNAME_CLAIM=preferred_username
     volumes:
       - ./:/app
-    command: /bin/sh -c "wait-for-it.sh -t 60 db:5432 -- ./manage.py migrate && ./manage.py runserver 0.0.0.0:80"
+    command: /bin/sh -c "wait-for-it.sh -t 60 db:5432 -- ./manage.py migrate && ./manage.py serve --port 80 --req-queue-len=200"
+    ports:
+      - "81:81"
```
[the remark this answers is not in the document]
fugal-dy: I was considering this so the probes are exposed in dev too, for verification. it doesn't reallly give great advantage over the native django dev server though, I agree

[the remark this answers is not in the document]
fugal-dy: I'd prefer to have it there, too though, otherwise the `make start` command would behave differently, which would be unexpected

winged:
Did you just supersede this with #893?

cmd.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/sh
 
-GUNICORN_WORKERS="${GUNICORN_WORKERS:-8}"
-
 ./manage.py collectstatic --noinput
 
-wait-for-it.sh "${DJANGO_DATABASE_HOST}":"${DJANGO_DATABASE_PORT}" -t "${WAITFORIT_TIMEOUT}" -- \
-    ./manage.py migrate --no-input && \
-    gunicorn --workers=$GUNICORN_WORKERS --bind=0.0.0.0:80 timed.wsgi:application
+wait-for-it.sh "${DJANGO_DATABASE_HOST}":"${DJANGO_DATABASE_PORT}" -t "${WAITFORIT_TIMEOUT}" --
```
winged, line 5: We already have a shell script. I think it'd be a good point in time to rewrite it in a more readable way.

You can put `set -x` on top of the script to ensure we don't continue in case of failure. (Also see our Shell coding guidelines for further info on flags etc)

docker-compose.override.yml
```diff
@@ -14,7 +14,9 @@ services:
       - DJANGO_OIDC_USERNAME_CLAIM=preferred_username
     volumes:
       - ./:/app
-    command: /bin/sh -c "wait-for-it.sh -t 60 db:5432 -- ./manage.py migrate && ./manage.py runserver 0.0.0.0:80"
+    command: /bin/sh -c "wait-for-it.sh -t 60 db:5432 -- ./manage.py migrate && ./manage.py serve --port 80 --req-queue-len=${HURRICANE_REQ_QUEUE_LEN:-250} --autoreload --static"
```
winged, line 17: Why not just run the `cmd.sh` as it's basically doing the same apart from the `--autoreload --static` flags, which could easily be added if in "dev" mode

winged (commented):
Looks good, but I suggest it's time to make the entrypoints / shell scripts a bit more readable

PR #896, opened 2022-10-18
fix(container): executable bit for cmd.sh

hairmare:
The container is currently failing to start in CI due to missing executable rights. This adds the executable bit to the `cmd.sh` entrypoint again.

* https://github.com/adfinis/helm-charts/pull/828#issuecomment-1282123405


winged (approved):
Nice catch!

PR #900, opened 2022-10-27
chore: add Make commands for easier development

trowik:
Added some commands to make the development of the app a little bit easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,26 +1,57 @@
-.PHONY: help start stop test shell flush loaddata
 .DEFAULT_GOAL := help
 
+.PHONY: help
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort -k 1,1 | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+.PHONY: start
 start: ## Start the development server
 	@docker-compose up -d --build
 
+.PHONY: stop
 stop: ## Stop the development server
 	@docker-compose stop
 
-test: ## Test the project
-	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8 && poetry run pytest --no-cov-on-fail --cov"
+.PHONY: lint
+lint: ## Lint the project
+	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8"
 
-shell: ## Shell into the backend
+.PHONY: format-code
+format-code: ## Format the backend code
+	@docker-compose exec backend sh -c "poetry run black . && poetry run isort ."
+
+.PHONY: test
+test: ## Test the project
+	@docker-compose exec backend sh -c "poetry run black . && poetry run isort . && poetry run pytest --no-cov-on-fail --cov
+
+.PHONY: bash
+bash: ## Shell into the backend
 	@docker-compose exec backend bash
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,26 +1,57 @@
...
+	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8"
+
+.PHONY: format-code
+format-code: ## Format the backend code
+	@docker-compose exec backend sh -c "poetry run black . && poetry run isort ."
+
+.PHONY: test
 test: ## Test the project
-	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8 && poetry run pytest --no-cov-on-fail --cov"
+	@docker-compose exec backend poetry run pytest --no-cov-on-fail --cov
```
open-dynaMIX, line 25: YMMV, but I like it if `make test` also lints the code.

timed/settings.py
```diff
@@ -61,6 +61,7 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
     "django.contrib.staticfiles",
     "rest_framework",
     "django_filters",
+    "django_extensions",
```
open-dynaMIX, line 64: `django-extensions` is a dev dependency (and should be), so this is not gonna work in prod.

You should add a block like this below `INSTALLED_APPS`:

```python
if ENV == "dev":
    INSTALLED_APPS.append("django_extensions")
```

open-dynaMIX (changes requested):

Makefile
trowik: Ok, changed it back

timed/settings.py
trowik: Done. Thanks a lot.

Makefile
```diff
@@ -1,26 +1,57 @@
...
+	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8"
+
+.PHONY: format-code
+format-code: ## Format the backend code
+	@docker-compose exec backend sh -c "poetry run black . && poetry run isort ."
+
+.PHONY: test
 test: ## Test the project
-	@docker-compose exec backend sh -c "poetry run black --check . && poetry run flake8 && poetry run pytest --no-cov-on-fail --cov"
+	@docker-compose exec backend sh -c "poetry run black . && poetry run isort . && poetry run pytest --no-cov-on-fail --cov
```
open-dynaMIX, line 25:
```suggestion
	@docker-compose exec backend sh -c "poetry run black . && poetry run flake8 . && poetry run pytest --no-cov-on-fail --cov
```

timed/settings.py
```diff
@@ -74,6 +74,9 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
     "timed.subscription",
 ]
 
+if ENV == "dev":
```
open-dynaMIX, line 77: This is really set to `dev` in dev?

open-dynaMIX (changes requested):
Some more nitpicks / questions ;)

timed/settings.py
trowik: Yes, it's written in `README.md`

>## Development

To get the application working locally for development, make sure to create a file `.env` with the following content:

```
ENV=dev
DJANGO_OIDC_CREATE_USER=True
````

open-dynaMIX: :+1:

open-dynaMIX (approved):

PR #908, opened 2022-11-23
feat(redmine): update expenditures on redmine projects

trowik:
added a command to update expenditures such as amount offered, amount invoiced project estimated hours and project spent hours for redmine projects


fugal-dy (commented):
looking good

PR #911, opened 2022-11-29
Redmine timed migration

trowik:
Update amount offered and invoiced for each project based on values from redmine.
This migration will be run once and removed afterwards.


fugal-dy (approved):
Well done. Looks good

PR #916, opened 2022-12-14
Revert "refactor(app-server): replace uWSGI with Django Hurricane"

trowik:
This reverts commit c7c14f016cc238831602f280bcd0e335337b4313.
Just in case we run into issues with the deployment of Django Hurricane.


fugal-dy (approved):
Looking good

cmd.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/sh
 
-# All parameters to the script are appended as arguments to `manage.py serve`
+GUNICORN_WORKERS="${GUNICORN_WORKERS:-8}"
 
 set -x
 
-./manage.py manage.py collectstatic --noinput
```
fugal-dy, line 7: this should probably also be changed in master, too
:grimacing:

PR #918, opened 2023-01-04
fix(reports): fix project and customer statistics

trowik:
fix ordering for project and customer statistics
include customer serializer for project statistics
add tests for filtering customer and project statistics


winged (approved):
Nice work, looks good!

PR #919, opened 2023-01-12
fix(tracking): fix absence for users with multiple employments

trowik:
1. fix: move the PublicHoliday exclusion from absence object manager   
to AbsenceViewSet get_queryset()

2. fix: only exclude public holidays from the current employment.
This was an issue for users with multiple employments (past and current) at different locations,
as locations may have different public holidays.


fugal-dy:
Reviewed together. Looks good!

fugal-dy (approved):

PR #922, opened 2023-01-16
fix(tracking): fix remaining effort check when creating report

trowik:
Check if remaining effort tracking is active on the corresponding project when creating a new report as well.
Previously the check only worked when updating a report, but not when creating one (no instance when creating a new report).


fugal-dy (approved):
All good

PR #923, opened 2023-01-19
fix: add missing rejected field to ReportIntersectionSerializer

trowik:
Just added missing field `rejected` to the Serializer


fugal-dy (approved):
LGTM

PR #925, opened 2023-01-23
fix(auth): failing requests result in status_code 401

### LINKED ISSUE #924: bug(auth): requests to the api with an invalid token receive a response status 500 instead of 401

fugal-dy:
closes #924

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: isort
         name: isort
         language: system
-        entry: isort -y
+        entry: isort
         types: [python]
       - id: flake8
         name: flake8
--- a/timed/authentication.py
+++ b/timed/authentication.py
@@ -8,6 +8,7 @@ from django.core.cache import cache
 from django.core.exceptions import SuspiciousOperation
 from django.utils.encoding import force_bytes
 from mozilla_django_oidc.auth import LOGGER, OIDCAuthenticationBackend
+from rest_framework.exceptions import AuthenticationFailed
 
 
 class TimedOIDCAuthenticationBackend(OIDCAuthenticationBackend):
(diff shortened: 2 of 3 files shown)


timed/authentication.py
```diff
@@ -33,22 +34,34 @@ def get_introspection(self, access_token, id_token, payload):
...
-            if "client_id" not in claims:
-                raise SuspiciousOperation("client_id not present in introspection")
+        while True:
+            try:
+                claims = self.cached_request(
+                    self.get_userinfo, access_token, "auth.userinfo"
+                )
+                break
+            except requests.HTTPError as e:
+                if not (e.response.status_code in [401, 403]):
```
winged, line 44:
```suggestion
                if e.response.status_code not in [401, 403]:
```

timed/authentication.py
```diff
@@ -33,22 +34,34 @@ def get_introspection(self, access_token, id_token, payload):
...
-            ):
-                raise e
-
-            # check introspection if userinfo fails (confidental client)
-            claims = self.cached_request(
-                self.get_introspection, access_token, "auth.introspection"
-            )
-            if "client_id" not in claims:
-                raise SuspiciousOperation("client_id not present in introspection")
+        while True:
```
winged, line 37: Why the loop? There is no condition where the code ever *actually* loops

timed/authentication.py
```diff
@@ -33,22 +34,34 @@ def get_introspection(self, access_token, id_token, payload):
...
+            try:
+                claims = self.cached_request(
+                    self.get_userinfo, access_token, "auth.userinfo"
+                )
+                break
```
winged, line 42: you could just early-return here
```suggestion
                return claims
```

timed/authentication.py
```diff
@@ -33,22 +34,34 @@ def get_introspection(self, access_token, id_token, payload):
...
+                    try:
+                        # check introspection if userinfo fails (confidential client)
+                        claims = self.cached_request(
+                            self.get_introspection, access_token, "auth.introspection"
+                        )
+                        if "client_id" not in claims:
+                            raise SuspiciousOperation(
+                                "client_id not present in introspection"
+                            )
+                        break
```
winged, line 56:
```suggestion
                        return claims
```

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
       - id: isort
         name: isort
         language: system
-        entry: isort -y
+        entry: isort
```
winged, line 12: Unrelated change - can you split it into a separate commit?

winged (changes requested):

timed/authentication.py
fugal-dy: Actually to avoid returning early ;)

and only once.

Changed it.

.pre-commit-config.yaml
fugal-dy: i did.

winged (approved):

PR #928, opened 2023-02-09
fix(tracking): allow null values on remaining effort for reports

trowik:


derrabauke (approved):
LGTM! Release the ~~dragon~~ fix! :dragon:

PR #931, opened 2023-02-14
fix(tracking): fix setting of remaining effort

trowik:
Change signal to pre_save and update most_recent_remaining_effort on task and total_remaining_effort on project if the value of remaining_effort on report has changed.

### DIFF
--- a/timed/tracking/signals.py
+++ b/timed/tracking/signals.py
@@ -1,5 +1,5 @@
 from django.db.models import Sum
-from django.db.models.signals import post_save, pre_save
+from django.db.models.signals import pre_save
 from django.dispatch import receiver
 
 from timed.tracking.models import Report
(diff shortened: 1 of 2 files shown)


timed/tracking/signals.py
```diff
@@ -15,20 +15,28 @@ def update_rejected_on_reports(sender, instance, **kwargs):
...
-            .get("sum_remaining")
-        )
-        project.total_remaining_effort = total_remaining_effort
-        project.save()
+    # ignore signal when loading a fixture
+    if kwargs.get("raw", False):  # pragma: no cover
+        return
+
+    if instance.pk:
+        if instance.task.project.remaining_effort_tracking:
```
fugal-dy, line 26: It's more agreeable when reading code if unmet conditions are ruled out immediately aka free up memory where possible and stay a close to the left border as possible.

```suggestion
    if not instance.pk:
        return
    if instance.task.project.remaining_effort_tracking is not True:
        return
```

timed/tracking/tests/test_report.py
```diff
@@ -1954,3 +1954,71 @@ def test_report_remaining_effort_total(
...
+
+    url = reverse("report-detail", args=[report_2.id])
+
+    response = internal_employee_client.patch(url, data)
+    assert response.status_code == status.HTTP_200_OK
+
+    report.task.refresh_from_db()
+    assert report.task.most_recent_remaining_effort == timedelta(hours=3)
+    assert report.task.project.total_remaining_effort == timedelta(hours=3)
+
```
fugal-dy, line 2006: maybe some comment along the line that the most recent remaining effort value should only be updated if and only if the `remaining_effort` property has changed on the report instance (and not blindly take the latest report that might have changed something else completely)

fugal-dy (approved):
good job =)

timed/tracking/signals.py
trowik: makes sense, thanks :)

timed/tracking/tests/test_report.py
trowik: Updated the doc string in the signal

PR #932, opened 2023-02-15
feat(redmine): add pretend mode to redmine commands

trowik:
Add pretend mode for update_project_expenditure and import_project_data commands so they can be tested without using productive data on test systems.

### DIFF
--- a/timed/redmine/management/commands/update_project_expenditure.py
+++ b/timed/redmine/management/commands/update_project_expenditure.py
@@ -9,6 +9,13 @@ from timed.projects.models import Project
 class Command(BaseCommand):
     help = "Update expenditures on associated Redmine projects."
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--pretend",
+            action="store_true",
+            help="Pretend mode for testing",
+        )
+
     def handle(self, *args, **options):
         redmine = redminelib.Redmine(
             settings.REDMINE_URL,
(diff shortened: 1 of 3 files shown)


timed/redmine/management/commands/update_project_expenditure.py
```diff
@@ -9,6 +9,15 @@
...
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--pretend",
+            dest="pretend",
+            default=True,
+            help="Pretend mode for testing",
+            type=bool,
+        )
+
```
fugal-dy, line 20: Add an argument as a switch with the action `store_true` or `store_false` (depending on the desired default value)

```suggestion
        parser.add_argument(
            "--pretend",
            action="store_true",
            help="Pretend mode for testing"
        )

```

timed/redmine/management/commands/update_project_expenditure.py
```diff
@@ -65,11 +76,18 @@ def handle(self, *args, **options):
...
+            if not pretend:
+                try:
+                    issue.save()
+                except redminelib.exceptions.BaseRedmineError as e:  # pragma: no cover
+                    self.stdout.write(
+                        self.style.ERROR(
+                            f"Failed to save Project {project.name} with Redmine issue {issue.id}!\n{e}"
+                        )
                     )
+
```
fugal-dy, line 88: `continue` missing?

If this is not pretending and throws an exception, it will both write an error and success message, won't it?

trowik: Yep, you're right. Thanks :)

fugal-dy (approved):
:100:

PR #938, opened 2023-02-20
fix(tracking): fix remaining effort on report creation

### LINKED ISSUE #933: bug: remaining effort on initial report creation is buggy
## Scenario
Create a report and add an estimated remaining effort to it.

## Expected behavior
It updates the remaining effort of the certain task in the statistics view.

## Actual behavior
After submitting the report the remaining effort does **not** update in stats view. But after editing the rem. effort and saving the report again in the timesheet, it will update the values in the stats view.

The attributes are fine in both requests to the backend. I suspect that a the `POST` and `PATCH` requests have different handling of the remaining effort attribute. Maybe the `POST` does not properly trigger the update?
image

trowik:
update most_recent_remaining_effort on task and total_remaining_effort on project when creating a report

Fixes #933

### DIFF
--- a/timed/tracking/signals.py
+++ b/timed/tracking/signals.py
@@ -26,21 +26,30 @@ def update_most_recent_remaining_effort(sender, instance, **kwargs):
     if kwargs.get("raw", False):  # pragma: no cover
         return
 
-    if not instance.pk:
-        return
     if instance.task.project.remaining_effort_tracking is not True:
         return
 
-    if instance.remaining_effort != Report.objects.get(id=instance.id).remaining_effort:
-        task = instance.task
-        task.most_recent_remaining_effort = instance.remaining_effort
-        task.save()
+    # update most_recent_remaining_effort and total_remaining_effort on report creation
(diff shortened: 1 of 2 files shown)


timed/tracking/signals.py
```diff
@@ -26,11 +26,23 @@ def update_most_recent_remaining_effort(sender, instance, **kwargs):
     if kwargs.get("raw", False):  # pragma: no cover
         return
 
-    if not instance.pk:
-        return
     if instance.task.project.remaining_effort_tracking is not True:
         return
 
+	# update most_recent_remaining_effort and total_remaining_effort on report creation
+    if not instance.pk:
```
winged, line 33: This contains quite a bit of duplicated code. Could this be simplified?

winged (changes requested):

timed/tracking/signals.py
trowik: Created a function for the duplicated  code

winged (approved):

PR #944, opened 2023-02-27
chore(reports): disable project and customer statistics

trowik:


hairmare (approved):
this is for sure better than killing the database with the queries that have a very high cost

PR #945, opened 2023-02-28
fix(tracking): fix report update notifactions

trowik:
Send notifcations about rejected reports only if rejected is True


trowik:
Will refactor when #946 is merged

derrabauke (approved):
LGTM

PR #947, opened 2023-02-28
Fix statistics

winged:
Rewrite the statistics code to generate much more performant SQL, and fix&cleanup some other issues on the way. You can review commit-by-commit so things make more sense.

### fix(statistics): refactor multiqs to use filtering aggregates

Filtering aggregates are much easier to manage, and can be constructed
using way simpler queries. This improves performance significantly.

The resulting query is now not a multi-level-nested-subquery that needs
to be joined in awkward ways and causing headaches to the PostgreSQL
query planner and optimizer.

Multi-minute queries now take a fraction of a second.

### feat(statistics): support ordering in new queryset wrapper

Also add proper allowed ordering fields to the viewsets. This
has been missing, so ordering didn't even work before

### chore(api): re-enable statistics views

They had horrible performance before so were disabled, but work properly
and fast now.


trowik:
Closed due to #947

trowik (changes requested):

trowik (approved):
Amazing work :tada:

PR #948, opened 2023-02-28
fix(pytest): ignore "invalid escape sequence" deprecation warning

winged:
This causes failures in code that we cannot control, so should be ignored


trowik (approved):
:+1:

winged:
This commit is also part of #947 so we can ignore the coverage issues here - they will be fixed in that PR as well

PR #949, opened 2023-02-28
feat: empty sums in correcr ordering

winged:
The statistics should report zero hours instead of NULL. This makes the ordering of data actually work as expected


trowik (approved):
:+1:

PR #952, opened 2023-03-02
chore(reject template): add further information to reject mail template

tongpu:
This should make it easier for people receiving a mail about their reports being rejected to understand which action to take next.


winged (approved):
Thank you!

PR #954, opened 2023-03-15
fix(redmine): fix value check for custom fields

trowik:
Check if the `value` is empty, not if the `dict` is empty.

### DIFF
--- a/timed/redmine/management/commands/import_project_data.py
+++ b/timed/redmine/management/commands/import_project_data.py
@@ -58,14 +58,10 @@ class Command(BaseCommand):  # pragma: no cover
             )
 
             timed_project.amount_offered = (
-                amount_offered["value"]
-                if amount_offered != ""
-                else timed_project.amount_offered
+                amount_offered.get("value") or timed_project.amount_offered
             )
             timed_project.amount_invoiced = (
-                amount_invoiced["value"]
-                if amount_invoiced != ""
-                else timed_project.amount_invoiced
+                amount_invoiced.get("value") or timed_project.amount_invoiced
             )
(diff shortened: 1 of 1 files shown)


timed/redmine/management/commands/import_project_data.py
```diff
@@ -58,13 +58,13 @@ def handle(self, *args, **options):
             )
 
             timed_project.amount_offered = (
-                amount_offered["value"]
-                if amount_offered != ""
+                amount_offered.get("value")
+                if amount_offered.get("value")
                 else timed_project.amount_offered
```
winged, line 63:
```suggestion
                amount_offered.get("value") or timed_project.amount_offered
```

timed/redmine/management/commands/import_project_data.py
```diff
@@ -58,13 +58,13 @@ def handle(self, *args, **options):
...
             )
             timed_project.amount_invoiced = (
-                amount_invoiced["value"]
-                if amount_invoiced != ""
+                amount_invoiced.get("value")
+                if amount_invoiced.get("value")
                 else timed_project.amount_invoiced
```
winged, line 68:
```suggestion
                amount_invoiced.get("value") or timed_project.amount_invoiced
```

winged (changes requested):
Small readability / optimisation suggestion

timed/redmine/management/commands/import_project_data.py
trowik: ah yes, of course
thanks :)

winged (approved):

PR #959, opened 2023-04-03
fix(redmine): log estimated_hours in update_project_expenditure command

trowik:
Only added additional log statement


derrabauke (approved):
:+1:

PR #961, opened 2023-04-04
chore(docker): install wait-for-it from debian package

trowik:


derrabauke (approved):
Looks way cleaner than the pinned commit. :+1:

winged (approved):
YAS!

PR #964, opened 2023-05-03
fix(tracking): don't send emails when updating own rejected reports

### LINKED ISSUE #963: editing rejected report results in weird mail
# Issue
At first I had a Report actually rejected and got this mail:
```
Some of your reports have been rejected by a reviewer. Please get in contact
with them to clarify the reports. Most likely, you will just need to move
the reports to the correct project / task.

Reviewer: <Actualy Reviewer>


Date: mm/dd/yyyy
Duration: 7:00 (h:mm)
Task: x > y > z
Comment: <comment>
---
```
Then after i edited the Report i got this mail:
```
Some of your reports have been rejected by a reviewer. Please get in contact
with them to clarify the reports. Most likely, you will just need to move
the reports to the correct project / task.

Reviewer: <Myself>


Date: mm/dd/yyyy
Duration: 7:00 (h:mm)
Task: x > y > z
Comment: <edited comment>
---
```
After every edit i got another mail, where I am marked as the reviewer.

trowik:
Don't send emails when updating anything besides task on a rejected report that the user owns.
Updating the task on a rejected report should still automatically unreject the report.

Fixes #963

### DIFF
--- a/timed/tracking/views.py
+++ b/timed/tracking/views.py
@@ -165,16 +165,18 @@ class ReportViewSet(ModelViewSet):
         serializer = self.get_serializer(instance, data=request.data, partial=partial)
         serializer.is_valid(raise_exception=True)
 
-        fields = {
-            key: value
-            for key, value in serializer.validated_data.items()
-            # value equal None means do not touch
-            if value is not None
-        }
-        if fields and request.user != instance.user:
-            tasks.notify_user_changed_report(instance, fields, request.user)
-        if fields.get("rejected"):
-            tasks.notify_user_rejected_report(instance, request.user)
+        if request.user != instance.user:
+            # send a notification only when the user is updating someone else's report
+            fields = {
+                key: value
+                for key, value in serializer.validated_data.items()
+                # value equal None means do not touch
+                if value is not None
+            }
+            if fields:
+                tasks.notify_user_changed_report(instance, fields, request.user)
+            if fields.get("rejected"):
+                tasks.notify_user_rejected_report(instance, request.user)
 
         return super().update(request, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


trowik:
> I can see that this fixes the weirdness of being notified on rejected reports by oneself. One can still reject one's own reports, though. Is that an intended use case?

Fair point. I don't see an use case for a reviewer to reject his own report.

timed/tracking/views.py
```diff
@@ -173,7 +173,7 @@ def update(self, request, *args, **kwargs):
         }
         if fields and request.user != instance.user:
             tasks.notify_user_changed_report(instance, fields, request.user)
-        if fields.get("rejected"):
+        if fields.get("rejected") and request.user != instance.user:
             tasks.notify_user_rejected_report(instance, request.user)
```
fugal-dy, line 177: If notifications are never sent in case where the user does stuff to own reports, that could be made more explicit by writing all related to notification below a `if request.user != instance.user`

timed/tracking/views.py
```diff
@@ -171,10 +171,11 @@ def update(self, request, *args, **kwargs):
             # value equal None means do not touch
             if value is not None
         }
-        if fields and request.user != instance.user:
-            tasks.notify_user_changed_report(instance, fields, request.user)
-        if fields.get("rejected"):
-            tasks.notify_user_rejected_report(instance, request.user)
```
fugal-dy, line 177: you could even include the lines L168-173 and add a comment on the reason why this is done to avoid head scratching and guessing next time someone reads those lines

fugal-dy (approved):
nice job!

PR #965, opened 2023-05-08
chore(deps): update django to 3.2.19

trowik:


open-dynaMIX (changes requested):
Change looks good.

I think the commit message could be a bit better. This commit updates all the dependencies and pins django to `^3.2.19`.

open-dynaMIX (approved):

pyproject.toml
```diff
@@ -45,6 +45,7 @@ python-redmine = "^2.3.0"
 sentry-sdk = "^1.9.5"
 whitenoise = "^6.2.0"
 django-hurricane = "^1.3.4"
+openpyxl = "3.0.10"
```
fugal-dy, line 48: add some explanation and, if you have it, the link to an issue or something so we know when to unpin

trowik: Found the core issue, I guess. Added a comment and an explanation in the commit.

open-dynaMIX: Might also make sense to link the issue as done [here](https://github.com/adfinis/mySAGW/blob/main/api/requirements-base.txt#L15).

trowik: ah nice, there's already an issue for that.
Thanks a lot

PR #981, opened 2023-07-18
Added Employment change model, refactoring working calculation

amchangama:


timed/employment/models.py
```diff
@@ -259,54 +260,11 @@ def calculate_worktime(self, start, end):
         :returns:     tuple of 3 values reported, expected and delta in given
                       time frame
         """
```
fugal-dy, line 262: The docstring does not seem to fit the function anymore since you moved the code to another function that is called by it.

I wonder if it would be sufficient to create the part for the employment.

But I'm not familiar enough with the model situation here e. g. why the calculation is bound to the Location model in the first place..

fugal-dy (commented):
Nice contribution!

It looks like you have not blacked or linted all of the code.

I pointed out some things that struck me first off. Feel free to address them or not.

PR #988, opened 2024-02-29
fix: update report billed flag to be like project billed flag

MitanOmar:


MitanOmar:
fix https://github.com/adfinis/timed-frontend/issues/1090

timed/tracking/views.py
```diff
@@ -267,8 +267,7 @@ def bulk(self, request):
         if "task" in fields:
             # unreject report if task has changed
             fields["rejected"] = False
-            if fields["task"].project.billed:
-                fields["billed"] = fields["task"].project.billed
+            fields["billed"] = fields["task"].project.billed
```
fugal-dy, line 270: This could set `fields["billed"]` to `None` which is not the same as `False`. Not sure if it would be appropriate to default `fields["billed"]` to `False`, similar to the above `fields["rejected"]`

MitanOmar: i make it sure that the value will be true or false

fugal-dy (approved):

timed/tracking/views.py
```diff
@@ -267,8 +267,7 @@ def bulk(self, request):
         if "task" in fields:
             # unreject report if task has changed
             fields["rejected"] = False
-            if fields["task"].project.billed:
-                fields["billed"] = fields["task"].project.billed
+            fields["billed"] = True if fields["task"].project.billed else False
```
c0rydoras, line 270:
```suggestion
            fields["billed"] = fields["task"].project.billed
```

this should also work since its a boolean

c0rydoras (changes requested):

timed/tracking/views.py
hairmare: @c0rydoras that's where this started off at https://github.com/adfinis/timed-backend/pull/988#discussion_r1507771956

c0rydoras:
```py
fields["billed"] = bool(fields["task"].project.billed)
```

c0rydoras: if it isn't a boolean this accomplishes the same

PR #992, opened 2024-04-12
use ruff

c0rydoras:


hairmare:
don't forget to adapt the pre-commit config

PR #1045, opened 2024-04-16
refactor(tests): use factory-boy fixtures instead of importing factories

c0rydoras:


hairmare:
Should this also update pytest-factoryboy given that the update seems possible with these changes? wdyt?

c0rydoras:
it still has the exact same problems :c

hairmare:
> it still has the exact same problems :c

meh

hairmare (approved):

PR #1049, opened 2024-04-16
Ruff config

c0rydoras:


timed/authentication.py
```diff
@@ -14,11 +14,8 @@
 class TimedOIDCAuthenticationBackend(OIDCAuthenticationBackend):
     def get_introspection(self, access_token, id_token, payload):
         """Return user details dictionary."""
-
         basic = base64.b64encode(
-            f"{settings.OIDC_RP_INTROSPECT_CLIENT_ID}:{settings.OIDC_RP_INTROSPECT_CLIENT_SECRET}".encode(
-                "utf-8"
-            )
+            f"{settings.OIDC_RP_INTROSPECT_CLIENT_ID}:{settings.OIDC_RP_INTROSPECT_CLIENT_SECRET}".encode()
```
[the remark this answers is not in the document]
c0rydoras: i'm not sure, ruff just removed "utf-8" because thats the default

timed/employment/views.py
```diff
@@ -174,7 +174,7 @@ def _extract_date(self):
                 return datetime.datetime.strptime(pk.split("_")[1], "%Y-%m-%d")
 
             except (ValueError, TypeError, IndexError):
-                raise exceptions.NotFound()
+                raise exceptions.NotFound from None
```
hairmare, line 177: Shouldn't these raise from the original exception? It feels like we are throwing away readily available info here.

```python
            except (ValueError, TypeError, IndexError) as ex:
                raise exceptions.NotFound from ex
```

c0rydoras: i'm not sure

c0rydoras: i've adjusted it now

pyproject.toml
```diff
@@ -100,6 +100,7 @@ omit = [
     "*/apps.py",
     "*/admin.py",
     "manage.py",
+    "timed/redmine/management/commands/import_project_data.py",
```
hairmare, line 103: baye a `# TODO` might make sense here

(yes i'm aware that more `todo:` culture would also be required for this to help....)

let me think some and i'll turn this into a new issue and resolve it here

ruff.toml
```diff
@@ -11,27 +11,95 @@ docstring-code-format = true
...
+	"S311",    # factories may use pseudo-random generators
 ]
 
 [lint.isort]
 known-first-party = ["timed"]
 known-third-party = ["pytest_factoryboy"]
 combine-as-imports = true
+
+[lint.flake8-annotations]
+ignore-fully-untyped = true
```
hairmare, line 105: i'm on the fence on this one. is it in here for the long haul or do we aim at getting rid of it in future?

timed/authentication.py
```diff
@@ -29,42 +26,39 @@ def get_introspection(self, access_token, id_token, payload):
             verify=settings.OIDC_VERIFY_SSL,
             headers=headers,
             data={"token": access_token},
+            timeout=10,
```
hairmare, line 29: if we ever reach this, then o boy are we sol

in the long run there might be potential as to providing lower order primaries lie retry-backoff implementation that would help alleviate why we need to set a timeout now

ruff.toml
c0rydoras: I was thinking of it as temporary

hairmare: perfect!

timed/reports/views.py
```diff
@@ -31,18 +31,18 @@ class YearStatisticViewSet(AggregateQuerysetMixin, ReadOnlyModelViewSet):
...
+    )
 
     def get_queryset(self):
         queryset = Report.objects.all()
         queryset = queryset.annotate(year=ExtractYear("date")).values("year")
         queryset = queryset.annotate(duration=Sum("duration"))
-        queryset = queryset.annotate(pk=F("year"))
-
-        return queryset
+        return queryset.annotate(pk=F("year"))
```
winged, line 45: Old pattern was more readable IMHO, consider disabling the rule, as it's not adding anything of value

timed/settings.py
```diff
@@ -11,7 +12,7 @@
 django_root = environ.Path(__file__) - 2
 
 ENV_FILE = env.str("DJANGO_ENV_FILE", default=django_root(".env"))
-if os.path.exists(ENV_FILE):  # pragma: no cover
+if Path(ENV_FILE):  # pragma: no cover
```
winged, line 15: I think this is an antipattern. At least call `.exists()` on it, so your intention is clear when reading the code

```suggestion
if Path(ENV_FILE).exists():  # pragma: no cover
```

pyproject.toml
c0rydoras: yea, AFAIK that file wasn't picked up by pytest previously

timed/settings.py
c0rydoras: i actually forgot the .exists there

winged: And I thought you knew something I didn't - that `bool(Path("non-existing-file"))` might magically imply an existance check. Glad I caught this, because it doesn't. `bool(Path("non-existing")) is True`

timed/reports/filters.py
```diff
@@ -72,29 +70,25 @@ def filter_cost_center(self, queryset, name, value):
...
     def filter_queryset(self, queryset):
         qs = super().filter_queryset(queryset)
 
         duration_ref = self._refs["reports_ref"] + "__duration"
 
-        full_qs = qs._base.annotate(
+        full_qs = qs._base.annotate(  # noqa: SLF001
             duration=Coalesce(
-                Sum(duration_ref, filter=qs._agg_filters),
+                Sum(duration_ref, filter=qs._agg_filters),  # noqa: SLF001
```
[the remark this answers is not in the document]
c0rydoras: SLF001 is now disabled

timed/projects/migrations/0017_alter_billingtype_options_alter_costcenter_options_and_more.py
```diff
@@ -0,0 +1,37 @@
...
+        ),
+        migrations.AlterModelOptions(
+            name='task',
+            options={'ordering': ('name',)},
+        ),
+        migrations.AlterModelOptions(
+            name='tasktemplate',
+            options={'ordering': ('name',)},
+        ),
+    ]
```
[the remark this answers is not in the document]
c0rydoras: I edited the older migrations to use tuples and dropped this one

timed/employment/serializers.py
```diff
@@ -20,7 +21,7 @@
 
 
 class UserSerializer(ModelSerializer):
-    included_serializers = {
+    included_serializers: ClassVar = {
```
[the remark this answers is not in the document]
c0rydoras: I dropped this rule (RUF012)

timed/conftest.py
```diff
@@ -25,7 +25,7 @@ def register_module(module):
 register_module(tracking_factories)
 
 
-@pytest.fixture
+@pytest.fixture()
```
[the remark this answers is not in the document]
c0rydoras: it now enforces not having parentheses

timed/reports/filters.py
```diff
@@ -50,8 +49,7 @@ def filter_cost_center(self, queryset, name, value):
         * When viewing the statistic over tasks, only the tasks
           are filtered
         """
-
-        # TODO Discuss: Is this the desired behaviour by our users?
+        # TODO: Discuss Is this the desired behaviour by our users?  # noqa: TD002,TD003
```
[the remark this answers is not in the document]
hairmare: the todo linter wants us to add a responsible person's username and provide a link to an issue

we might also just consider satisfying it where applicable.

in this concrete instance the todo looks like left-overs from pre-merge times and not really like an actual todo tho

timed/reports/filters.py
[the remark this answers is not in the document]
hairmare: [SLF001](https://docs.astral.sh/ruff/rules/private-member-access/) feels like a good rule tho, are we sure that _agg_filters really should be private?

timed/employment/views.py
```diff
@@ -92,20 +92,21 @@ def get_queryset(self):
                     Q(task__in=assigned_tasks) | Q(user=user)
                 )
                 return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
-            raise exceptions.PermissionDenied("User has no employment")
+            msg = "User has no employment"
+            raise exceptions.PermissionDenied(msg) from exc
```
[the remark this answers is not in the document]
hairmare: can you elaborate on why breaking the exception chain makes sense here?

timed/employment/admin.py
```diff
@@ -51,12 +51,12 @@ class SuperviseeForm(forms.ModelForm):
...
         """Meta information for the supervisee form."""
 
-        fields = "__all__"
+        fields = "__all__"  # noqa: DJ007
         model = models.User.supervisors.through
 
 
 class SupervisorInline(admin.TabularInline):
-    autocomplete_fields = ["to_user"]
+    autocomplete_fields = ("to_user",)
```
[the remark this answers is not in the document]
c0rydoras: since i disabled `RUF012` it wouldn't complain about a list

timed/employment/admin.py
```diff
@@ -146,77 +144,77 @@ class AbsenceCreditInline(admin.TabularInline):
...
-    inlines = [
+    inlines = (
         SupervisorInline,
         SuperviseeInline,
         EmploymentInline,
         OvertimeCreditInline,
         AbsenceCreditInline,
-    ]
+    )
     list_display = ("username", "first_name", "last_name", "is_staff", "is_active")
```
hairmare, line 154: this unchanged line is worth mentioning in the context of https://github.com/adfinis/timed-backend/pull/1049/files#r1568378503

we should be consistent, either tuple all the things or list all the things

tbh, i'm not completely sure what django does with these under the hood, but their semantics don't seem to have any impact on the correctness of the resulting app

c0rydoras:
I'm not sure how to handle

> [!WARNING]
> DJ001 Avoid using `null=True` on string-based fields such as `CharField`

It would require a migration (if the null=True) gets dropped and might break something (don't know for sure tho)

timed/employment/admin.py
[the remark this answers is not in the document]
hairmare: been reading up on [RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/) do we have any mutability requirements on all of these class vars?

timed/settings.py
hairmare: wait, do we need to learn how write new ruff rules for this :grin:

hairmare:
> I'm not sure how to handle
> 
> Warning
> 
> DJ001 Avoid using `null=True` on string-based fields such as `CharField`
> 
> It would require a migration (if the null=True) gets dropped and might break something (don't know for sure tho)

I think i agree with [DJ001](https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/) per se as well as with your reasoning why this might need more consideration.

Let's deactivate this one for now and defer it to future us with a new issue.

timed/employment/serializers.py
```diff
@@ -285,7 +282,7 @@ class Meta:
         """Meta information for the location serializer."""
 
         model = models.Location
-        fields = ["name", "workdays"]
+        fields = ("name", "workdays")
```
hairmare, line 285: we could hint all of these oneliners to be mulltilined by `ruff format`
```suggestion
        fields = ("name", "workdays",)
```

This would reduce future churn by simplifying diffs at the cost of making a already huge file huger

This comment is on the smallest edge case of 2 fields, on other serializers further down the trade off is clearer

timed/reports/views.py
```diff
@@ -81,10 +81,11 @@ def __init__(self, catch_prefixes, *args, base_qs=None, agg_filters=None, **kwar
     def filter(self, *args, **kwargs):
         if args:  # pragma: no cover
             # This is a check against programming errors, no need to test
-            raise RuntimeError(
+            msg = (
                 "Unable to detect statistics filter type form Q objects. use "
                 "filter_aggregate() or filter_base() instead"
             )
+            raise RuntimeError(msg)
         my_filters = {
```
hairmare, line 89: This is a check against programming errors, no need to test 0.0

We could also update the signature to not allow this mistake
```python
    def filter(self, *, **kwargs):
        # no check here cause it's not needed and the simplified signature still enforced a kw only method
        my_filters = {
        ...
```

timed/employment/tests/test_employment.py
```diff
@@ -136,7 +136,8 @@ def test_employment_list_supervisor(auth_client):
     assert len(json["data"]) == 2
 
 
-def test_employment_unique_active(db):
+@pytest.mark.django_db()
+def test_employment_unique_active():
```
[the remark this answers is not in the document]
hairmare: Previously `db` was defined in the methods scope and never accessed, so using a mark does indeed do a tiny bit less while making it more obvious that some shenanigans happen in the complex db "fixture".

The alternative would be to figure out how to care about [ANN001](https://docs.astral.sh/ruff/rules/missing-type-function-argument/) and [ARG001](https://docs.astral.sh/ruff/rules/unused-function-argument/) down the line.

timed/employment/admin.py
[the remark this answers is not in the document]
hairmare: oh my, the [docs](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/) are unclear... for autocomplete_fields it's a list, but other fields like readonly_fields can be both.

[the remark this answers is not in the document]
hairmare: and, for autocomplete_fields, it looks like the default implementation returns `()` as a default but calls it a list in docs

so what we really will want this to be is `tuple(str, ...)`. this would make it immutable, which is a good thing, and also makes it clear that it is not a tuple containing variable types

it seems that typeddjango [agrees with my analysis](https://github.com/typeddjango/django-stubs/blob/e11955f05994e196dcee004811ce2e05ff230462/django-stubs/contrib/admin/options.pyi#L89) allowing both tuples and lists for the autocomplete_fields case, hence a `tuple(str, ...)` being the mostest correctest

timed/employment/serializers.py
```diff
@@ -29,7 +29,7 @@ class Meta:
         """Meta information for the user serializer."""
 
         model = get_user_model()
-        fields = [
+        fields = (
```
hairmare, line 32: most likely we will want introduce this kind of change in a later type hinting focused pr
```suggestion
        fields: tuple(str, ...) = (
```
even more likely there is room for a proper type hierarchy and i need to study the typeddjango efforts to know more

timed/employment/admin.py
winged: Nope, and as you noted elsewhere, DRF isn't too consistent either. I think they're doing tuples to avoid accidental mutations, but I try to avoid tuples in the general case as they can be a source of nasty errors, especially in the single-value kind - forget the trailing comma, and the braces become just braces, and if the containing value happens to be an iterable, and the "user" of the value happens to not explicitly type-check, you suddenly have  "f", "o", "o" instead of "foo" somewhere

winged:
> > I'm not sure how to handle
> > Warning
> > DJ001 Avoid using `null=True` on string-based fields such as `CharField`
> > It would require a migration (if the null=True) gets dropped and might break something (don't know for sure tho)
> 
> I think i agree with [DJ001](https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/) per se as well as with your reasoning why this might need more consideration.
> 
> Let's deactivate this one for now and defer it to future us with a new issue.
> 
> edit: link for TODO comment: #1052

That's DB design dogma however. I do agree with it in the general case, but I think that check just goes too far. As you noted, we can do it in a separate PR sometimes, but just for the linter's sake, it's not the right thing to do

timed/reports/filters.py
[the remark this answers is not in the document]
winged: Yeah generally I think it's alright - but then again, we can get into deep OOP philosophy discussions here:

Are private methods/properties considered private as in the classical OOP, "only accessible within the same class", or would it make sense to expand it as "should only be used by the same feature", where the feature may span across a few places outside of the original class... 🤔

In Python, private just means "please don't touch", not "you cannot, technically, have access to this".

I'm now swinging back and thinking we should enable the rule, but keep the code `#noqa`'d. And think of the rule as such: It's a marker to say "I know what I'm doing" (which was the case when I wrote the code here), but have the linter yell at you if you just happen to mess with private stuff without thinking too much about it

timed/reports/filters.py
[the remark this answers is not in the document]
winged: I think that's actually a good idea, as otherwise, TODOs do get lost over time. Just the other day I was stumbling over a TODO from over 12 years ago :grin: 

BUT: Given that Timed-backend is due to be merged with the frontend repo, the issue links will likely all change, thus I'd suggest disabling this rule for now

c0rydoras:
timed/subscription/admin.py:27:38: S324 Probable use of insecure hash functions in `hashlib`: `md5`
   |
25 |         password = self.cleaned_data.get("password")
26 |         if password is not None:
27 |             self.instance.password = hashlib.md5(password.encode()).hexdigest()
   |                                      ^^^^^^^^^^^ S324
28 |         return super().save(commit=commit)
   |

```py
class CustomerPassword(models.Model):
    """Password per customer used for login into SySupport portal.

    Password are only hashed with md5. This model will be obsolete
    once customer center will go live.
    """
```

Can this be dropped, respectively ignored and dropped in a separate PR?

Since https://github.com/adfinis/customer-center does exist

timed/reports/filters.py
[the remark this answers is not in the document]
c0rydoras: they have been disabled

timed/reports/filters.py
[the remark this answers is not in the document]
c0rydoras: i re-enabled it

timed/employment/serializers.py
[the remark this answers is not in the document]
hairmare: ClassVar provides value by making linters generate a warning during type checking
```python
us = UserSerialzer()
us.included_serializers = { ... } # this generates a warning or error with ClassVar
```
If this pattern is allowed, then the second line would override the classvar for all instances and not just `us`.

timed/reports/views.py
c0rydoras: AFAIK this isn't valid python code, maybe you meant this?
```py
def filter(self, /, **kwargs):
  ...
```

c0rydoras: i've implemented this change here: https://github.com/adfinis/timed-backend/pull/1049/commits/0425fddbdf73e76f1b62e6baac671e263d4e92cb

trowik:
> > timed/subscription/admin.py:27:38: S324 Probable use of insecure hash functions in `hashlib`: `md5` | 25 | password = self.cleaned_data.get("password") 26 | if password is not None: 27 | self.instance.password = hashlib.md5(password.encode()).hexdigest() | ^^^^^^^^^^^ S324 28 | return super().save(commit=commit) |
> > ```python
> > class CustomerPassword(models.Model):
> >     """Password per customer used for login into SySupport portal.
> > 
> >     Password are only hashed with md5. This model will be obsolete
> >     once customer center will go live.
> >     """
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can this be dropped, respectively ignored and dropped in a separate PR?
> > Since https://github.com/adfinis/customer-center does exist
> 
> I'm not 100% if this is still in use or not. It looks like it was part of the old SySupport portal and might have been replaced with a proper solution when the portal was migrated to being an Ember based frontend.
> 
> If this is in fact the case then we should create an issue to track the codes removal. Maybe @winged or @trowik know if it's still needed?

I'm also not 100% sure, but I'd say it's safe to remove it in another PR.

timed/reports/filters.py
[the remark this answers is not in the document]
hairmare: I like using noqa as a hint that we gave this proper thought.

IMO part of a classes contract when using a private variable is also "i wont tell you when i delete this and noticing+fixing this will be your issue". given it's (hopefully) only use is in this codebase this is probably fine

timed/employment/admin.py
hairmare: I didn't check DRF itself given that this is a Django Admin module. From the looks of it and given it's age the `django.contrib.admin` isn't where i'm looking for inspiration on how to write pythonic code.

timed/reports/views.py
hairmare: indeed, `*` gobbles args and kwargs while `/` just gobbles args

ruff.toml
hairmare: Let's create a followup issue for introducing more typing. While i don't expect us to fully type the code, there are nevertheless several lowhanging fruit where some simple type definitions will improve the code base and help us get rid of errors in several categories.

hairmare (approved):
i think this is shaping up nicely and we should proceed with merging it while deferring further improvements to later PRs.

As it stands, this already activates many rules that will help keep the code base clean going forward as well as stomping several potential bugs.

timed/employment/serializers.py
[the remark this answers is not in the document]
c0rydoras: I have re-enabled RUF012 and annotated all included_serializers as ClassVar[dict[str, str]]

timed/employment/serializers.py
c0rydoras: i've added trailing commas to all tuples i came across now

timed/employment/views.py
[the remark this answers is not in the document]
c0rydoras: Changed here: https://github.com/adfinis/timed-backend/pull/1049/commits/2dbcee4ba893671359666ba5c822c0da4156c544

timed/employment/admin.py
hairmare: The "f". "o", "o" bug is just as much taken care of by ruff format btw:

```console
$ cat test.py 
v = ("foo")
hairmare@lina:~/Documents/git.repos$ ruff format test.py 
1 file reformatted
hairmare@lina:~/Documents/git.repos$ cat test.py 
v = "foo"
$ vim test.py # hack hack hack
$ cat test.py 
v = ("foo",)
hairmare@lina:~/Documents/git.repos$ ruff format test.py 
1 file left unchanged
```

together with some simple typing as per #1054 i don't think this bug can be triggered w/o being caught in review

pyproject.toml
hairmare: just noticed that it's in `timed/redmine` let's consider it to be obsolecent soon and then this is fine as is

timed/employment/views.py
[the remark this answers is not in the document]
hairmare: `from None` is really just satisfying the linter with no benefits at all, imo keeping an intact dependency chain is the real winner on how to solve this

let's touch on this later if need be given it's just bikeshedding at this point.

timed/employment/serializers.py
hairmare: neat! i think this will help alleviate the "f", "o", "o" issue with single value tuple even further while introducing consistency

timed/reports/views.py
winged: Well, one could as well go without intermediate variables at all, which would be even cleaner (and not violate the rule):

```py
return (
    queryset.annotate()
        .annotate()
        .annotate()
        .annotate()....
)
```

hairmare:
btw, wrt to us not noticing that we forgot that `Path()` is just as truthy as `Path().exists()` i'm following up in ruff: https://github.com/astral-sh/ruff/issues/11048

timed/employment/admin.py
```diff
@@ -146,77 +144,87 @@ class AbsenceCreditInline(admin.TabularInline):
...
     disable_staff_status.short_description = _("Disable staff status of selected users")
 
-    def enable_staff_status(self, request, queryset):
+    def enable_staff_status(self, _, queryset):
         queryset.update(is_staff=True)
 
     enable_staff_status.short_description = _("Enable staff status of selected users")
 
-    def has_delete_permission(self, request, obj=None):
+    def has_delete_permission(self, _, obj=None):
```
winged, line 196: Instead of just taking `_` as parameter, you can use `_request` to denote that it's unused, but still document that it's actually a request. Alternatively, type hints

timed/employment/serializers.py
```diff
@@ -18,9 +21,12 @@
 from timed.employment import models
 from timed.tracking.models import Absence, Report
 
+if TYPE_CHECKING:
+    from typing import ClassVar
+
```
winged, line 26: Why is this needed? That looks like an ugly workaround... but for what?

timed/employment/views.py
```diff
@@ -92,20 +96,21 @@ def get_queryset(self):
                     Q(task__in=assigned_tasks) | Q(user=user)
                 )
                 return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
-            raise exceptions.PermissionDenied("User has no employment")
+            msg = "User has no employment"
+            raise exceptions.PermissionDenied(msg) from None
+        else:
+            return queryset
```
winged, line 102: This is weird control flow. This should be in the `try` block, where it belongs to the `if` on line 70, as it was before.

The way it is now makes the code unneccessarily hard to follow

timed/permissions.py
```diff
@@ -12,69 +11,68 @@
 class IsUnverified(BasePermission):
     """Allows access only to verified objects."""
 
-    def has_object_permission(self, request, view, obj):
+    def has_object_permission(self, _, __, obj):
```
winged, line 14:
```suggestion
    def has_object_permission(self, _request, _view, obj):
```

timed/projects/filters.py
```diff
@@ -33,7 +36,7 @@ class ProjectFilterSet(FilterSet):
     has_reviewer = NumberFilter(method="filter_has_reviewer")
     customer = NumberInFilter(field_name="customer")
 
-    def filter_has_manager(self, queryset, name, value):
+    def filter_has_manager(self, queryset, _, value):
```
winged, line 39:
```suggestion
    def filter_has_manager(self, queryset, _name, value):
```

timed/projects/filters.py
```diff
@@ -49,7 +52,7 @@ def filter_has_manager(self, queryset, name, value):
             )
         )
 
-    def filter_has_reviewer(self, queryset, name, value):
+    def filter_has_reviewer(self, queryset, _, value):
```
winged, line 55:
```suggestion
    def filter_has_reviewer(self, queryset, _name, value):
```

timed/projects/filters.py
```diff
@@ -69,16 +72,24 @@ class Meta:
...
     """Filter most frequently used tasks.
 
-    TODO:
+    Todo:
+    ----
     From an api and framework standpoint instead of an additional filter it
     would be more desirable to assign an ordering field frecency and to
     limit by use paging.  This is way harder to implement therefore on hold.
+
     """
```
winged, line 93: I think TODO or similar comments that dont *describe* the code should be comments, not docstrings

```suggestion
    """
    # TODO: From an api and framework standpoint instead of an additional filter it
    # would be more desirable to assign an ordering field frecency and to
    # limit by use paging.  This is way harder to implement therefore on hold.
```

timed/tracking/models.py
```diff
@@ -103,6 +103,19 @@ class Report(models.Model):
...
+    class Meta:
+        """Meta information for the report model."""
+
+        indexes = (models.Index(fields=["date"]),)
+
+    def __str__(self):
+        """Represent the model as a string.
+
+        :return: The string representation
+        :rtype:  str
```
winged, line 115: That docstring style could be cleaned up while we're at it. sphinx-doc I think used this format, but we don't do that anymore

winged (changes requested):

timed/employment/serializers.py
c0rydoras: Its not evaluated at runtime

timed/employment/admin.py
c0rydoras: i didn't know this were possible, its a lot more readable

timed/employment/views.py
c0rydoras: https://docs.astral.sh/ruff/rules/try-consider-else/

timed/projects/filters.py
c0rydoras: changed here: https://github.com/adfinis/timed-backend/pull/1049/commits/91a42864c398905d969df1272ba681346a5b15a1

timed/projects/filters.py
c0rydoras: changed here: https://github.com/adfinis/timed-backend/pull/1049/commits/91a42864c398905d969df1272ba681346a5b15a1

timed/permissions.py
c0rydoras: changed here: https://github.com/adfinis/timed-backend/pull/1049/commits/91a42864c398905d969df1272ba681346a5b15a1

timed/tracking/models.py
c0rydoras: i moved all `__str__`s and some others (clean, validate on serializers, worktime_per_day, duration etc,) to type annotations here: https://github.com/adfinis/timed-backend/pull/1049/commits/78cd741caf2add4c1bde1c12065a5d0a2ae445d2

The rest i would like to do when adding types (https://github.com/adfinis/timed-backend/issues/1054#issuecomment-2068946433)

timed/employment/serializers.py
c0rydoras: this is especially useful to circumvent circular imports when typing things, in this case its not really necessary

timed/employment/views.py
winged: Yeah one can discuss it in the simple case as in the docs, but here you're still returning *sometimes* from the `try` block, which this rule doesn't catch. You're basically moving a non-exception-related bit of code to the exception structure, which makes things confusing.

Looking at the code again - what you *should* do is to move all the code out of the try/except that doesn't actually need to be there.

```python
        try:
            current_employment = models.Employment.objects.get_at(
                user=user, date=datetime.date.today()
            )
        except models.Employment.DoesNotExist:
            if CustomerAssignee.objects.filter(user=user, is_customer=True).exists():
                assigned_tasks = Task.objects.filter(
                    Q(
                        project__customer__customer_assignees__user=user,
                        project__customer__customer_assignees__is_customer=True,
                    )
                )
                visible_reports = Report.objects.all().filter(
                    Q(task__in=assigned_tasks) | Q(user=user)
                )
                return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
            msg = "User has no employment"
            raise exceptions.PermissionDenied(msg) from None

        if current_employment.is_external:
            assigned_tasks = Task.objects.filter(
                Q(task_assignees__user=user, task_assignees__is_reviewer=True)
                | Q(
                    project__project_assignees__user=user,
                    project__project_assignees__is_reviewer=True,
                )
                | Q(
                    project__customer__customer_assignees__user=user,
                    project__customer__customer_assignees__is_reviewer=True,
                )
            )
            visible_reports = Report.objects.all().filter(
                Q(task__in=assigned_tasks) | Q(user=user)
            )
            return queryset.filter(Q(reports__in=visible_reports) | Q(id=user.id))
    else:
        return queryset

```

timed/employment/serializers.py
winged: Typing allows referencing other types by string representation, which I prefer instead of this weird dance (That isn't even neccessary here)

c0rydoras: AFAIK, when using string references you would still import things in a `TYPE_CHECKING` block

timed/authentication.py
```diff
@@ -29,42 +26,39 @@ def get_introspection(self, access_token, id_token, payload):
...
                     if e.response.status_code not in [401, 403]:  # pragma: no cover
-                        raise e
-            raise AuthenticationFailed()
+                        raise
+                else:
+                    return claims
+            raise AuthenticationFailed from exc
 
-    def get_or_create_user(self, access_token, id_token, payload):
+    def get_or_create_user(self, access_token, *args, **kwargs):
```
winged, line 60: I've thought a good bit about this change (and similar ones elsewhere). I'm now of the opinion that we should keep the old named params, as they're part of the *interface*, even if not used in this particular implementation. Instead either disable the offending rule or do a #noqa marker. 

`*args, **kwargs` is for when you don't know what you might get as parameters, and possibly passing it on, but we *do* know the parameters, we just don't use them. Therefore IMNSHO they should be kept as before

timed/conftest.py
```diff
@@ -14,8 +14,8 @@
 
 
 def register_module(module):
-    for name, obj in inspect.getmembers(module):
-        if isinstance(obj, FactoryMetaClass) and not obj._meta.abstract:
+    for _name, obj in inspect.getmembers(module):
```
winged, line 17: Is `name` a reserved word now? or future reserved word? IMO there's no reason to rename this var

timed/employment/serializers.py
winged: This still breaks and goes against all intuition, as `ClassVar` is referenced just below. In my own tests, this breaks even if `not TYPE_CHECKING`

winged (changes requested):

timed/conftest.py
c0rydoras: AFAIK no, but its unused and ruff doesn't like that, an alternative would be `_`

c0rydoras:
https://github.com/adfinis/timed-backend/pull/1049#discussion_r1576161714

@winged this only works with

```py
from __future__ import annotations
```

See [PEP 563](https://peps.python.org/pep-0563/)

timed/conftest.py
winged: Ah right, then leave it as is, my bad

winged:
> [#1049 (comment)](https://github.com/adfinis/timed-backend/pull/1049#discussion_r1576161714)
> 
> @winged this only works with
> 
> ```python
> from __future__ import annotations
> ```
> 
> See [PEP 563](https://peps.python.org/pep-0563/)

TIL!

But I think that __future__ thingy is more about forward declarations, not about not having to import stuff that you *can* import. Do weird things only if you MUST, we shouldn't blindly follow dogma

timed/authentication.py
c0rydoras: i changed them back, respectively used `_{argname}` instead of `{argname}` where possible

https://github.com/adfinis/timed-backend/pull/1049/commits/717820e0c781b888aafa76ff0605e50ab7f92378

timed/employment/views.py
hairmare: this is so nice, while the ruff check didn't point us towards the exact direction of a proper fix. the rule helped surface a code smell that we were able to remediate, leading to a more correct implementation in this method

i think that is quite neat :mechanical_arm:

hairmare:
> TIL!

Um, same :)

I promised @c0rydoras that i wold do some light catching up on PEPs wrt to typing, specifically wrt to the previously unknown to me string form, and i discovered a rabbit hole of PEPs that i have yet to go down fully having previously focused on getting started with typing and not deep typing. 

I still have some reading to do, while i'm not (yet?) convinced that the string for is pythonic, i do think `TYPE_CHECKING` has value beyond just being dogma. tbh, i only figured this out when using it.

Basically, the import section has long been a part of my code that i do inspect, but i rarely write it on my own. either my IDE just add the required imports or i write them in any (isort violating)  order and ruff fixes whatever i am too lazy to consider at the time of writing. Given this hand-off approach to handling imports, the `TYPE_CHECKING` block seems to help me have a deeper understanding of the code i write.

I've run into situations where i refactor the code and a dependency was "ruffed" to move my import into the block, this helps demonstrate that i was successful at reducing the coupling between the component under refactor and what i was working on.

Anyways.... @c0rydoras proposed to today that he would love to merge this in the near term and followup on typing discussions in #1054 and I kinda agree with him. 

I think this discussion is highly valuable and don't consider it bike-shedding at all. In my opinion we can start safely using some typing without the resulting code being highly un-pythonic. I created #1054 because i see considerable value where typing can help us get rid of several low-level  incorrectness's in our code base.

For this PR to get merged ASAP, which i am highly favorable of since it seems to be correct and not introduce and defects i think we have two options:
* Remove any disputable typing related changes and follow up discussing them in #1054 
* Merge them as is and follow up discussing them in #1054

To no surprise i'm in favour of the latter option. The plan is to tackle #1054 right once this is landed.

We should also prepare the typing discussion so we can find consensus/input with a broader audience of stake holders to any policy here.

langer rede kurzer sinn, :pleading_face: :point_right: :point_left: 

does this address your concerns?

winged:
I absolutely agree with you @hairmare - The thing I'm most scared about with this strict Ruff config is that we stop questioning the rules in cases where they don't make sense, and apply patterns just to satisfy the linter.

We've seen this a few times in this PR (not blaming @c0rydoras at all): It's often difficult to see if a linter complaint is a valid criticism or something that should be acceptable in this particular situation.

So in a weird sense having stricter linting solves a problem but introduces a new one: While earlier you had to rely on experience/knowledge to figure out whether a code structure/pattern/whatever is "good", you now have to rely on experience/knowledge/common-sense to figure out whether a linter complaint is "good". Not sure which one is preferrable. For example the whole tuple-vs-list discussion we had here: I think tuples are harder to read and easier to make mistakes with, for the (slight) benefit of them being readonly (but still replaceable)

Regarding type hints - I think it's a glorious idea, and it makes my life (via IDE/LSP/tooling) much better, but: Python will never be a strictly, fully-typed language, so I'm quite happy to make the move rather gradual, and omit it where it's just cluttering up the code without real benefit.

So, in conclusion:

* More, stricter linters are good, but do not replace common sense / experience.
* Type hints are good, but don't be dogmatic about it


I'm OK with merging this PR soon-ish - I'll have a final look over it and approve if I don't see any more issues

timed/authentication.py
```diff
@@ -12,13 +12,10 @@
 
 
 class TimedOIDCAuthenticationBackend(OIDCAuthenticationBackend):
-    def get_introspection(self, access_token, id_token, payload):
+    def get_introspection(self, access_token, _id_token, _payload):
```
winged, line 15: Just a note (no change required): This naming pattern really only works when the method is called with positional args. We'll have to be careful to check the call sites in such cases - especially if it's implementing a protocol / defined API or overriding a base class' implementation

timed/projects/filters.py
```diff
@@ -107,9 +114,7 @@ def filter(self, qs, value):
         )
         qs = qs.annotate(frequency=Count("reports")).order_by("-frequency")
         # limit number of results to given value
-        qs = qs[: int(value)]
-
-        return qs
+        return qs[: int(value)]
```
winged, line 117: This whole filter is kind of smelly - should be cleaned up in a separate PR:

* It's not a filter only
* It does implicit ordering
* It does implicit "pagination"

To make it work properly, this will need corresponding fixes in the frontend as well I think

winged (approved):
Nice work! There's still a few code smells, but they're either from before your change (and so should not be addressed here) or can be fixed / cleaned up in a follow-up PR.

I'd say you have green light to merge this from me

winged:
> @winged how would you merge this?
> 
> I'd be in favour of squashing it and then using the `fix` scope

You have my approval, and @hairmare's as well - go ahead :+1:

PR #1050, opened 2024-04-17
chore(ci): semantic release

c0rydoras:


.github/workflows/release.yaml
```diff
@@ -1,21 +1,40 @@
...
 
+      - name: Semantic Release
+        uses: go-semantic-release/action@v1
+        id: semrel
+        with:
+          github-token: ${{ secrets.ADFINISBOT_PAT }}
+
+      - name: Setup Poetry
+        if: steps.semrel.outputs.version != ''
+        uses: abatilo/actions-poetry@v3.0.0
```
hairmare, line 28: did you consider just running `pipx install poetry` instead of this action?

c0rydoras: i adjusted it to use pipx

.github/workflows/release.yaml
```diff
@@ -30,19 +45,20 @@ jobs:
...
 
       - name: Build and push
-        id: docker_build_ghcr
+        if: steps.semrel.outputs.version != ''
         uses: docker/build-push-action@v5
         with:
           context: .
           file: ./Dockerfile
-          push: ${{ github.event_name != 'pull_request' }}
+          push: true
```
hairmare, line 61: maybe it could run on git push without doing any actual pushing?
```suggestion
          push: ${{ steps.semrel.outputs.version != '' }}
```
We probably don't need this since building the dev image for testing covers this already?

c0rydoras: the prod image also gets built in the test.yaml file

.github/workflows/release.yaml
```diff
@@ -1,21 +1,36 @@
 ---
-name: Release Container Image
+name: Release
 
 on:
   push:
     branches: [main]
-    tags:
-      - "v*.*.*"
+  workflow_call:
```
hairmare, line 7: i already runs on main after we merge this. when would we need this manual trigger?

c0rydoras: i'm not sure, copy pasted it from https://github.com/adfinis/outdated

c0rydoras: i removed it

hairmare (approved):

PR #1055, opened 2024-04-19
chore(dev-env): drop compose.override.yaml

c0rydoras:
there should be need for it because the compose setup is only for developement


hairmare:
can you add it to .gitignore as well?

hairmare (approved):

c0rydoras:
> can you add it to .gitignore as well?
done

PR #1056, opened 2024-04-19
code signing

c0rydoras:
chore(release): sign image using cosign
chore(release): create SBOM attestation and attach it to signed image
chore(trivy-scan): now has attest input
chore(schedule): now creates an attestation using cosign

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,6 +9,18 @@ jobs:
   semrel:
     name: Semantic Release
     runs-on: ubuntu-latest
+    permissions:
+      actions: none
+      checks: none
+      contents: none
+      deployments: none
+      issues: none
+      packages: write
+      pull-requests: none
+      repository-projects: none
+      security-events: none
+      statuses: none
+      id-token: write # needed for signing the images with GitHub OIDC using cosign
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -58,3 +71,25 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: |
             ${{ steps.meta.outputs.labels }}
+
+      - name: Run Trivy vulnerability scanner on api
+        if: steps.semrel.outputs.version != ''
+        uses: aquasecurity/trivy-action@0.19.0
```
hairmare, line 77: is there a reason you are not just calling the trivy-scan workflow?

c0rydoras: IIRC i had the same problem in https://github.com/adfinis/outdated and it needed the digest, additionally IIRC it would have to be an action to be callable as step

c0rydoras: If we can move this to an action that would be great, i briefly looked at composite actions and they look promising

hairmare: i think i also asked the same q in outdated :)

i guess this is ok for now and we can for sure investigate the possibility of an action in the future

hairmare (approved):

PR #1058, opened 2024-04-19
fix(semrel): correctly set version tags

c0rydoras:


.github/workflows/release.yaml
```diff
@@ -44,6 +44,10 @@ jobs:
           images: ghcr.io/adfinis/timed-backend
           flavor: |
             latest=auto
+          tags: |
+            type=semver,pattern={{version}},value=${{ steps.semrel.outputs.version }}
+            type=semver,pattern={{major}}.{{minor}},value=${{ steps.semrel.outputs.version }}
+            type=semver,pattern={{major}},value=${{ steps.semrel.outputs.version }}
```
hairmare, line 50: don't we also need {{major}}.{{minor}}.{{patch}}?

c0rydoras: thats what `version` is for AFAIK

c0rydoras: image
https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver

hairmare: tru dat

hairmare (approved):

Issue #1053, opened 2024-04-18
Drop CustomerPassword


c0rydoras:
[S324](https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/): 

> > timed/subscription/admin.py:27:38: S324 Probable use of insecure hash functions in `hashlib`: `md5` | 25 | password = self.cleaned_data.get("password") 26 | if password is not None: 27 | self.instance.password = hashlib.md5(password.encode()).hexdigest() | ^^^^^^^^^^^ S324 28 | return super().save(commit=commit) |
> > ```python
> > class CustomerPassword(models.Model):
> >     """Password per customer used for login into SySupport portal.
> > 
> >     Password are only hashed with md5. This model will be obsolete
> >     once customer center will go live.
> >     """
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can this be dropped, respectively ignored and dropped in a separate PR?
> > Since https://github.com/adfinis/customer-center does exist
> 
> I'm not 100% if this is still in use or not. It looks like it was part of the old SySupport portal and might have been replaced with a proper solution when the portal was migrated to being an Ember based frontend.
> 
> If this is in fact the case then we should create an issue to track the codes removal. Maybe @winged or @trowik know if it's still needed?

I'm also not 100% sure, but I'd say it's safe to remove it in another PR.

_Originally posted by @trowik in https://github.com/adfinis/timed-backend/issues/1049#issuecomment-2063371137_

PR #1062, opened 2024-04-22
fix: drop customerpassword

c0rydoras:


trowik:
Ref https://github.com/adfinis/timed-backend/issues/1053

trowik (approved):
Nice work

PR #1071, opened 2024-04-26
Type annotations

c0rydoras:


timed/authentication.py
```diff
@@ -10,9 +13,18 @@
...
+
+    from django.db.models import QuerySet
+
+    from timed.employment.models import User
+
 
 class TimedOIDCAuthenticationBackend(OIDCAuthenticationBackend):
-    def get_introspection(self, access_token, _id_token, _payload):
+    def get_introspection(
+        self, access_token: str, _id_token: str, _payload: dict
```
hairmare, line 26: I take it you're not a fan of `Self`?
```suggestion
        self: Self, access_token: str, _id_token: str, _payload: dict
```

timed/employment/models.py
```diff
@@ -244,14 +252,7 @@ def calculate_worktime(self, start, end):
         6. The absences are all absences for this user within time frame
         7. The balance is the reported time plus the absences plus the
             overtime credit minus the expected worktime
-
```
hairmare, line 247: I'd have to look at what standard y'all wanna use but afaict something like this could help us keep helpful comments
```
Parameters:
-----------------
start: calculate worktime starting on given day.
end:   calculate worktime till given day

Return:
----------
tuple of 3 values reported, expected and delta in given
````

@winged do you know if we decided on a new preferred docs format?

File: timed/reports/views.py [hunk 3]
```diff
@@ -296,7 +308,15 @@ def _generate_workreport_name(self, from_date, today, project):
             self._clean_filename(project.name),
         )
 
-    def _create_workreport(self, from_date, to_date, today, project, reports, user):  # noqa: PLR0913
+    def _create_workreport(  # noqa: PLR0913
+        self,
+        from_date: date,
+        to_date: date,
+        today: date,
```
hairmare, line 315: Passing today seems like something we'd want to take a closer look at and refactor in a diff PR.

timed/authentication.py
c0rydoras: Its mostly just not necessary, most IDEs respectively LSPs know how to handle `self` without an annotation

winged: Really not neccessary, as that's given by definition. `Self` can be used as return value or when another parameter of the same type is taken somewhere

timed/employment/models.py
winged: I honestly don't want to force anything. I think the rules in [PEP257](https://peps.python.org/pep-0257/) are just fine.

Only reason we had some weird syntax is so sphinx-doc (and for a while mypy as well I think) which we don't use anymore, could parse out the type hints. We don't need that anymore.

I'd suggest just using plain text description of the parameters as well as the return values. "Just make it clear what's going on" (and use wording according to PEP257 ofc)

winged (approved):
LGTM, but I think we should not spend too much time adding annotations to existing code. The added value gain is not *that* great to go over everything and "clean it up" all the way.

It's a good thing when writing new code, and I appreciate you doing it here, but let's not start doing this in all projects ;-)

timed/reports/views.py
c0rydoras: yup: #1074

timed/employment/models.py
c0rydoras: I've added back documentation about the return types:

https://github.com/adfinis/timed-backend/blob/4420bf0c089da08220a8d77c07425a4c7eedb2a2/timed/employment/models.py#L256

https://github.com/adfinis/timed-backend/blob/4420bf0c089da08220a8d77c07425a4c7eedb2a2/timed/employment/models.py#L378

timed/authentication.py
hairmare: the docs call it "harmless" and i agree with y'all

c0rydoras:
We could also consider using these:

- https://github.com/typeddjango/django-stubs
- https://github.com/typeddjango/djangorestframework-stubs

PR #1082, opened 2024-05-06
chore(dev-setup): fix wait4ports and permission issues on /var/www/st…

c0rydoras:
…atic

the permissions of /var/www/static can become an issue because we run as another user (usually 1000) so that linting/formatting doesn't mess up permissions


winged (approved):
LGTM

